PR #95, opened 2021-11-20
README: "Paland's printf" instead of "printf"

amakukha:
Paland's printf is quite popular in the embedded world. I just added his last name so that: 1) it's clearer what "printf" the README is referring to; 2) it's more googleable (e.g., when someone searches for "tinyprintf vs Paland printf")

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ nanoprintf is statically configurable so users can find a balance between size, 
 
 [tinyprintf](https://github.com/cjlano/tinyprintf) doesn't print floating point values.
 
-"[printf](https://github.com/mpaland/printf)" defines the actual standard library `printf` symbol, which isn't always what you want. It stores the final converted string (with padding and precision) in a temporary buffer, which makes supporting longer strings more costly. It also doesn't support the `%n` "write-back" specifier.
+Paland's "[printf](https://github.com/mpaland/printf)" defines the actual standard library `printf` symbol, which isn't always what you want. It stores the final converted string (with padding and precision) in a temporary buffer, which makes supporting longer strings more costly. It also doesn't support the `%n` "write-back" specifier.
 
 Also, no embedded-friendly printf projects that I could find are both in the public domain *and* have single-file implementations.
 


charlesnicholson (approved):

charlesnicholson:
Thanks for the clarification!

charlesnicholson:
BTW on the same day you opened this PR, @eyalroz emailed me asking if I'd update the text to point at his taking-over-and-fixing-up of the Paland printf repo, so I ended up updating the text away what you wrote. Do you two know each other? It's funny that two people would bring this up on the same day after months of quiet in the repo :)

amakukha:
Yeah, I asked him about nanoprintf on StackOverflow. Specifically, I was interested in which one produces smaller binary.

Now you answered my question in the README :) Thanks!

eyalroz:
@amakukha : Hi :-)

Anyway, nanoprintf takes the cake here, I think. I mean, I'm  no expert in  minimizing code size; and libprintf is pretty small, but still. There should probably be a more straight-up comparison of the sizes when the feature sets are configured to be identical (which in itself is not trivial).

amakukha:
@eyalroz , actually, in my MIPS32 project using nanoprintf instead of Paland's printf increases the binary size by 8.5 KB. Even though nanoprintf object is smaller by 8%. Not sure why.

eyalroz:
@amakukha : Interesting... like I said, this is not my expertise (I don't develop for very small devices - my original interest is for GPU use.

The phenomenon you describe may be due to `libprintf` being compiled with different inlining of the central "big" function, as different visible functions in the object, and if you only use one of them, the rest get discarded. Or - because of LTO of the compiled code which slims it down. But I'm just speculating.

Also, I think we should move this discussion either to email or to a new issue about how to properly compare compiled sizes.

charlesnicholson:
@amakukha Whoa, that's a surprise! Do you have a redacted mapfile you can share? Is the compiler you're using openly available? I'd love to take a look at what's going on.

To test the nanoprintf compiled size, I generally take https://github.com/charlesnicholson/nanoprintf/blob/main/examples/use_npf_directly/your_project_nanoprintf.cpp, set all the define flags to 0, then compile that file to an object with my target architecture and the "optimize for size" flag enabled. Then, I use nm with sort-by-size enabled and sum up the sizes of each symbol.

How are you measuring the 8.5KB bloat?

Also it occurred to me that if you're using the static visibility flag (e.g. https://github.com/charlesnicholson/nanoprintf/blob/main/tests/static_nanoprintf.c) in a header file, and then including that header file here and there, you will get a separate copy of nanoprintf in each c file that includes the header.

PR #125, opened 2021-12-26
Binary conversion support

charlesnicholson:
Add a "b" format specifier that prints the value's representation in base 2. Compatible with all unsigned modifiers. Alternate form prints "0b" prefix. 

I believe the implementation to be compliant with http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2630.pdf


charlesnicholson:
@dl8dtl FYI, thanks for your work on N2630! This PR adds what I believe to be a full implementation of binary printing to nanoprintf :)

dl8dtl:
> FYI, thanks for your work on N2630!

You're welcome.
My initial proposal was to finally add `0b` numeric literals to the C standard (all other major languages already had it), and during the discussion of that proposal, it was suggested that formatted binary IO would be a logical counterpart, so I volunteered for that proposal as well.

Issue #246, opened 2023-07-25
Wrong value for floats on atmega328p (arduino).


jimktrains:
Here is my example program. The `nanoprintf` directory is a `git clone` of this repository. 

```c++
typedef size_t ssize_t;

#define NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_BINARY_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS 0

// Compile nanoprintf in this translation unit.
#define NANOPRINTF_IMPLEMENTATION
#include "nanoprintf/nanoprintf.h"

int my_snprintf(char *buffer, size_t bufsz, char const *fmt, ...) {
  va_list val;
  va_start(val, fmt);
  int const rv = npf_vsnprintf(buffer, bufsz, fmt, val);
  va_end(val);
  return rv;
}


void setup() {
  Serial.begin(115200);
  float f = 5.4;
  char buf[15];
  my_snprintf(buf, 15, "% 6.2f", f);
  Serial.println(buf);
}

void loop() {
}
```

This gives:

```
 517.40
```

(The value is wrong, but it's also placing a leading space to make the whole thing 7 char long instead of 6?)

I have a test program I compiled on my laptop(`Linux lamb 5.15.0-76-generic #83-Ubuntu SMP Thu Jun 15 19:16:32 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux` via `g++ -o test test.cc`), which works fine.

```c++
#define NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_BINARY_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS 0

[...]

PR #247, opened 2023-07-25
Issue #246. Fixes formatting floats via gcc-avr.

jimktrains:
On platforms that have a 16-bit `unsigned int`, the `implicit_one` constant is mis-calculated. Explicitly casting this to a `uint32_t` fixes the issue.

Running tests via `./b` has all test passing.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -5,3 +5,4 @@
 * [Dino Dai Zovi](https://github.com/ddz)
 * [Shang Yuanchun](https://github.com/ideal)
 * [Shreyas Balakrishna](https://github.com/shreyasbharath)
+* [Jim Keener](https://github.com/jimktrains)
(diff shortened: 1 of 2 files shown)


jimktrains:
For some context. Floats seemed to be disabled on the Arduino/gcc-avr for sprintf and friends.

I can't find an exact documentation link, but it comes up in the forum.

https://forum.arduino.cc/t/sprintf-a-float-number/1013193/5
https://forum.arduino.cc/t/sprintf-with-float-values/937562/3

I'm sure there are better/smaller/less-intensive ways to handle this, but I'm just trying to get something working to work out some other things right now, and nanprintf is small enough that it's fitting with my application with room to spare.

charlesnicholson:
Ah, yeah, if it's using newlib-nano that has float formatting disabled by default :(

I'm happy to merge this as-is, but noticed that you didn't update the CONTRIBUTORS.md file. Just to make sure you didn't forget, I'll ask you to update it or confirm here that you don't want it updated, either is AOK w/me!

charlesnicholson (approved):

jimktrains:
I didn't think about it really. I'll add myself to it.

Issue #249, opened 2023-09-13
Document the return values


charlesnicholson:
Add a "return values" section to the README.md, and to the header file.

See https://github.com/charlesnicholson/nanoprintf/discussions/248 for discussion.

charlesnicholson:
added in https://github.com/charlesnicholson/nanoprintf/pull/250

PR #250, opened 2023-09-13
document return values

charlesnicholson:


charlesnicholson:
https://github.com/charlesnicholson/nanoprintf/issues/249

PR #253, opened 2024-01-10
Fix for accessing string content when precision is 0.

DeanoBurrito:
Hey there,
First of all thanks for nanoprintf - it's been a great addition to a few of my projects.

I came across an issue (or rather unexpected behaviour) when using it in my kernel, specifically when formatting a string as `%.*s` and passing 0 as the precision. As expected this will write 0 characters, but the pointer for the string content is still de-referenced in the loop condition (to check for a null terminator). This is fine if the pointer for the string content is valid, but if the pointer is null it results in null deref.

It might sound silly wanting to write a string with a precision of 0 and null for the data, but my use case is printing human-readable debug data passed from a device driver (where strings are represented with a data pointer + length). In this case a driver might not want to provide anything so it sets the length to 0 and data pointer to null. I could (and probably should) check for this across kernel/driver boundaries, but I feel it'd be nice to have this in nanoprintf as well.

Thanks,
Dean.

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -744,7 +744,7 @@ int npf_vpprintf(npf_putc pc, void *pc_ctx, char const *format, va_list args) {
         cbuf = va_arg(args, char *);
 #if NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS == 1
         for (char const *s = cbuf;
-             s && *s && ((fs.prec_opt == NPF_FMT_SPEC_OPT_NONE) || (cbuf_len < fs.prec));
+             ((fs.prec_opt == NPF_FMT_SPEC_OPT_NONE) || (cbuf_len < fs.prec)) && *s;
              ++s, ++cbuf_len);
 #else
         for (char const *s = cbuf; *s; ++s, ++cbuf_len); // strlen
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -744,7 +744,7 @@ int npf_vpprintf(npf_putc pc, void *pc_ctx, char const *format, va_list args) {
         cbuf = va_arg(args, char *);
 #if NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS == 1
         for (char const *s = cbuf;
-             *s && ((fs.prec_opt == NPF_FMT_SPEC_OPT_NONE) || (cbuf_len < fs.prec));
+             s && *s && ((fs.prec_opt == NPF_FMT_SPEC_OPT_NONE) || (cbuf_len < fs.prec));
              ++s, ++cbuf_len);
 #else
         for (char const *s = cbuf; *s; ++s, ++cbuf_len); // strlen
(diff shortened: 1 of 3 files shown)


charlesnicholson:
Heya @DeanoBurrito, thanks for the pull request and I think your use case makes good sense. I'll take a look ASAP!

nanoprintf.h
```diff
@@ -744,7 +744,7 @@ int npf_vpprintf(npf_putc pc, void *pc_ctx, char const *format, va_list args) {
         cbuf = va_arg(args, char *);
 #if NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS == 1
         for (char const *s = cbuf;
-             *s && ((fs.prec_opt == NPF_FMT_SPEC_OPT_NONE) || (cbuf_len < fs.prec));
+             s && *s && ((fs.prec_opt == NPF_FMT_SPEC_OPT_NONE) || (cbuf_len < fs.prec));
```
charlesnicholson, line 747: I think I'd prefer a rearrangement to a new clause:
```suggestion
             ((fs.prec_opt == NPF_FMT_SPEC_OPT_NONE) || (cbuf_len < fs.prec)) && *s;
```

That way the code footprint should stay the same, and this specific change stays minimal. I'd prefer to do "nanoprintf gracefully handles null strings" in a different PR after giving it some thought- we might want to print `<null>` or something similar.

charlesnicholson (commented):
I left a suggestion, also could I ask you to add a specific unit test case for this in `tests/unit_vpprintf.cc` please?

Alternatively I'm happy to do this in a new PR if you don't have time / appetite. If you choose to do it, though, consider adding yourself to the contributors file at the root as well!

nanoprintf.h
DeanoBurrito: yeah thats more elegant, I've made the change.

DeanoBurrito:
Test case is added! Let me know if there's any changes you'd like. 
I went with passing the precision of 0 in the format string directly (rather than as a vararg argument, like in my use case) in order to keep the test as simple as possible. Since handling star args is tested elsewhere and wasnt part of the issue.

charlesnicholson (approved):

charlesnicholson:
Looks great, thanks!

DeanoBurrito:
@charlesnicholson Hey again, do you want something done about the failing tests? They're all triggered by `-Werror=format-overflow` reporting the input to `%s` is null.

charlesnicholson:
I'd recommend adding something like 

```
  #pragma GCC diagnostic ignored "-Wformat-overflow"
```

near the top, around line 15-16. That should shut gcc up; it's very helpful except for when we want to do unsafe things intentionally to test :)

DeanoBurrito:
hmm, I'll give that a try - thanks :)

DeanoBurrito:
That appears to have done it, thanks for the support.

charlesnicholson:
Third time's the charm, I guess :) Thanks for the work!

Issue #251, opened 2023-10-18
npf_vsnprintf writes beyond bufsz


arjanoskam:
When calling npf_vsnprintf() in order to write a limited number of characters (bufsz parameter) into the destination buffer, the function writes beyond the bufsz offset into the destination buffer.
The issue was found when using npf_vsnprintf() to write small pieces (50 chars) of a larger text buffer (>500 chars) to a FIFO before sending it to a UART. No format specifiers, just characters.

I would expect the function to write no more than bufsz chars into the destination buffer, because that buffer could be limited in size (the very reason for calling this function).

Currently npf_vsnprintf() simply calls npf_vpprintf(), which does not seem to consider the size of the destination buffer and just processes the entire format string. The option of putting a terminating 0 at the beginning or end (bufsz-1) of the buffer in case of overflow is not sufficient, because the overflow has already occurred and the memory following the destination buffer is corrupted. Also the return value n is still the result of the entire format string, but it should be at most bufsz.

charlesnicholson:
Hi @arjanoskam, thanks for reporting this. `npf_vsnprintf` should behave as you expect, not writing more than `bufsz` bytes to the specified buffer. There's a unit test in `unit_snprintf.cc` that tests this behavior, but it's only one test and not exactly exhaustive:

```
  SUBCASE("doesn't write past end of buffer") {
    buf[3] = '*';
    REQUIRE(npf_snprintf(buf, 3, "abcd") == 4);
    REQUIRE(buf[0] == 'a');
    REQUIRE(buf[1] == 'b');
    REQUIRE(buf[2] == 'c');
    REQUIRE(buf[3] == '*');
  }
```

`npf_vsnprintf` does call `npf_vpprintf`, but the internal per-byte callback function is `npf_bufputc`, which (hopefully?) guards against buffer overrun. The string bounds are stored in an `npf_bufputc_ctx_t` struct that `npf_vsnprintf` creates on the stack and passes in as the "context" parameter:

```
void npf_bufputc(int c, void *ctx) {
  npf_bufputc_ctx_t *bpc = (npf_bufputc_ctx_t *)ctx;
  if (bpc->cur < bpc->len) { bpc->dst[bpc->cur++] = (char)c; }
}
```

So, of course it's possible that there's a bug here, but I'm not seeing it yet. If you can provide a simple repro case, I'd appreciate it and happily look into it.

charlesnicholson:
[...]

PR #256, opened 2024-01-13
snprintf truncate behavior by default

charlesnicholson:
`npf_snprintf` and `npf_vsnprintf` were incorrectly allowing strings to be non-terminated unless the `NANOPRINTF_SNPRINTF_SAFE_TRIM_STRING_ON_OVERFLOW` flag was specified. But, these functions should never allow unterminated strings; that's what the `n` is for.

This PR makes `..._TRIM_STRING_...` the default behavior for `npf_snprintf` and `npf_vsnprintf`.


charlesnicholson:
From the conversation in https://github.com/charlesnicholson/nanoprintf/issues/251

PR #265, opened 2024-06-16
Eliminate the signed integer conversion function

Okarss:
On my Cortex-M7 build this reduces the size by 64/160 B for normal/large specifiers respectively. And there is no undefined behavior involved:
https://stackoverflow.com/questions/50605/signed-to-unsigned-conversion-in-c-is-it-always-safe
https://stackoverflow.com/questions/17313579/is-there-a-safe-way-to-get-the-unsigned-absolute-value-of-a-signed-integer-with

MSVC throws an error for unary minus operator on unsigned types, so I made it into `0 - ...` expression instead.


charlesnicholson (approved):
I love it.

Okarss:
Introduced the `NPF_NOINLINE` because even with `Os` optimization GCC was inlining `npf_utoa_rev()` function and, as it is now called in 3 places, actually increasing the size. Moved the function parameter of `npf_uint_t` type to be the first one, because of the typical ABI "alignment" specifics for passing parameters of types larger than the CPU registers. That includes, for example, 64-bit types on 32-bit ARM and should generally be more optimal when using large format specifiers.

What really surprised me:
```
buf[n] = ... ;
++n;
```
```
buf[n++] = ... ;
```
These are compiled differently and the latter one tends to generate a larger binary...

nanoprintf.h
```diff
@@ -17,9 +17,15 @@
...
 #if defined(__clang__) || defined(__GNUC__) || defined(__GNUG__)
+  #define NPF_NOINLINE __attribute__((noinline))
   #define NPF_PRINTF_ATTR(FORMAT_INDEX, VARGS_INDEX) \
     __attribute__((format(printf, FORMAT_INDEX, VARGS_INDEX)))
 #else
+  #ifdef _MSC_VER
+    #define NPF_NOINLINE __declspec(noinline)
+  #else
+    #define NPF_NOINLINE
+  #endif
```
charlesnicholson, line 28: I try to keep this top part of the header as the "public" API, since this is where users look. Since `NPF_NOINLINE` is an implementation detail and not used in the public portion of the file, could I ask you to move it down to the implementation section please?

nanoprintf.h
```diff
@@ -894,9 +896,9 @@ int npf_vpprintf(npf_putc pc, void *pc_ctx, char const *format, va_list args) {
         } else
 #endif
         {
-          unsigned const base = (fs.conv_spec == NPF_FMT_SPEC_CONV_OCTAL) ?
+          uint_fast8_t const base = (fs.conv_spec == NPF_FMT_SPEC_CONV_OCTAL) ?
```
charlesnicholson, line 899: In previous iterations / compilers, I've found that using non-register-width types can incur an extra instruction penalty for masking off the upper unused bits (if any). Using native register sizes (which I wish C had a type for) let the compiler emit smaller code because it knew that these expression used the whole register width. Does this change cause an interesting change with respect to the generated assembly?

charlesnicholson (approved):

charlesnicholson:
> Introduced the `NPF_NOINLINE` because even with `Os` optimization GCC was inlining `npf_utoa_rev()` function and, as it is now called in 3 places, actually increasing the size. Moved the function parameter of `npf_uint_t` type to be the first one, because of the typical ABI "alignment" specifics for passing parameters of types larger than the CPU registers. That includes, for example, 64-bit types on 32-bit ARM and should generally be more optimal when using large format specifiers.
> 
> What really surprised me:
> 
> ```
> buf[n] = ... ;
> ++n;
> ```
> 
> ```
> buf[n++] = ... ;
> ```
> 
> These are compiled differently and the latter one tends to generate a larger binary...

Discovering micro-behaviors like this is so fascinating to me. Over in [nanocobs](https://github.com/charlesnicholson/nanocobs) I discovered that gcc (at the time) did significantly better with manipulating scalar array index variables instead of manipulating pointers directly. It's an infinite source of mystery for those of us unwilling to take the next step and start reading GIMPLE :)

nanoprintf.h
Okarss: Good point! I will move it.

nanoprintf.h
Okarss: That would be the case with `uint8_t` and `uint_least8_t`, but not the `uint_fast8_t`. Here is an [example](https://stackoverflow.com/a/59593614) of how the _fast_ `stdint.h` types are defined on different platforms. These types have been introduced exactly for this purpose - to be portable and optimal at the same time. Essentially they provide the functionality actually necessary in real life, contrary to the C's base types.

On 32-bit ARM the `uint_fast8_t` is actually the same 32-bit unsigned integer and doesn't change anything at all. Because the the `int` is not allowed to be smaller than 16-bit, the biggest impact is on CPUs smaller than that. The best popular example is the 8-bit AVR, but for the 16-bit MSP430 and RL78 the code also tends to be smaller. On AVR I basically can observe that for every place, where the 16-bit `int/unsigned/enum` is accessed, the binary size increases by additional 2, 4 or more bytes compared to 8-bit data types.

Okarss:
> It's an infinite source of mystery

Add to that different architectures, compilers, configuration settings... And the variations indeed go to infinity. :(

By the way, I have one more minor and similarly quick improvement in the making and will make a PR soon. After that I can see an overhaul of the integer types to use more of the `stdint.h` provided ones, but that will not be such a quick one...

charlesnicholson:
Sounds great- let me know when you're running out of PR ideas so I can cut a new release :)

Issue #260, opened 2024-03-21
Version in nanoprintf.h


MCMattiaOetiker:
Hi,

thank you a lot for your work.

I'm using nanoprintf at work by copying and pasting the nanoprintf.h file inside my cpp project.

I noticed that inside the nanoprintf.h the version is not indicated. If I open the file inside my project I don't know what version nanoprintf is.

Would it be possible to add the version in the comment section of the nanopritnf.h file for future releases?

Thanks in advance for your time

charlesnicholson:
Thanks for the suggestion; I think it's a great idea and I've almost done it a few times but have been lazy. I'll start doing it with the next release. :)

PR #266, opened 2024-06-19
Optimize format specifier parsing

Okarss:
This one reduces the number of assignment operations and does comparisons against `NPF_FMT_SPEC_OPT_NONE`, which equals zero. On AVR and PIC32 it manages to decrease the binary size by a whopping 96 and 128 bytes respectively. Though even more surprisingly it also somehow manages to increase the binary size by 16 bytes on my Cortex-M7 build.

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -221,6 +221,7 @@ typedef enum {
 } npf_format_spec_length_modifier_t;
 
 typedef enum {
+  NPF_FMT_SPEC_CONV_NONE,
   NPF_FMT_SPEC_CONV_PERCENT,      // '%'
   NPF_FMT_SPEC_CONV_CHAR,         // 'c'
   NPF_FMT_SPEC_CONV_STRING,       // 's'
@@ -244,24 +245,21 @@ typedef enum {
 } npf_format_spec_conversion_t;
 
 typedef struct npf_format_spec {
-  char prepend;          // ' ' or '+'
-  char alt_form;         // '#'
-
 #if NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS == 1
-  npf_fmt_spec_opt_t field_width_opt;
   int field_width;
+  uint8_t field_width_opt; // npf_fmt_spec_opt_t
   char left_justified;   // '-'
   char leading_zero_pad; // '0'
 #endif
-
 #if NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS == 1
-  npf_fmt_spec_opt_t prec_opt;
   int prec;
+  uint8_t prec_opt;        // npf_fmt_spec_opt_t
 #endif
-
-  npf_format_spec_length_modifier_t length_modifier;
-  npf_format_spec_conversion_t conv_spec;
-  char case_adjust;
+  char prepend;          // ' ' or '+'
+  char alt_form;         // '#'
+  char case_adjust;      // 'a' - 'A'
+  uint8_t length_modifier; // npf_format_spec_length_modifier_t
+  uint8_t conv_spec;       // npf_format_spec_conversion_t
 } npf_format_spec_t;
 
 #if NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS == 0
@@ -366,7 +364,7 @@ static int npf_parse_format_spec(char const *format, npf_format_spec_t *out_spec
   }
 #endif
 
-  int tmp_conv = -1;
+  uint_fast8_t tmp_conv = NPF_FMT_SPEC_CONV_NONE;
   out_spec->length_modifier = NPF_FMT_SPEC_LEN_MOD_NONE;
   switch (*cur++) { // Length modifier
     case 'h':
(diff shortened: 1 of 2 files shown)


Okarss:
So, after the second commit there is almost no change of binary size for CM0, tenths of bytes decrease for CM4, MSP430, RL78, x64, and an astonishing 274 B reduction for AVR. The stack size should also be reduced slightly for all platforms.

An interesting thing I found out for non-negative local variables used with `enum` constants - some compilers treat `enum` types as `char`, others as `int`. Overall the order of preference for types of such variables seems to be this (starting with the most preferred):
1. `uint_fast8_t`
2. `int_fast8_t`
3. `enum`
4. `int`

The downsides are that such a change looses the semantic context of type names (though I added the corresponding `enum` types in comments) and live debugging becomes a bit more cumbersome. Are you OK with this?

Okarss:
> let me know when you're running out of PR ideas so I can cut a new release :)

At this point there are two ideas:
- General overhaul of the integer types to use more of the `stdint.h` provided ones.
- Refactoring of the format specifier logic. But I'm still not even sure what exactly to do with it and whether it can improve anything.

Both of these will take a significant time to think through and test. Therefore in my opinion after this PR the `nanoprintf` deserves a release. :smile:

charlesnicholson:
I agree with your preference-order list, I think it makes good sense. The only reason I was using `int` and `unsigned` was because I didn't want to pay the masking / sign-extension cost, but you seem to have navigated that via the `fastX_t` types.

I'm comfortable fully removing the enumerated types and just using `uint_fast8_t` everywhere. Perhaps just making the enums nameless, and then assigning them to int types? To me that's preferable to `#define` values...
```
enum {
  NPF_FOO = ...,
  NPF_BAR = ...
};

// later
uint_fast8_t const x = NPF_BAR;
```

BTW the reason the format specifier parsing logic is as arcane as it is is because of all of the rules- many options nullify other options or combine to form yet other options, etc. I added that horrifying cascading fall-through thing in the switch statement because it was a way to add support for a corner-case rule without increasing the binary size on Cortex-M things... I'm sure there are less insane and more canonical ways to do these things, but it was a quick win at the time :)

Okarss:
Indeed the `enum` type names are no longer necessary. I will remove them. Maybe we should add a generic `// enum` comment at the end of the definitions of integer variables used for enumerations just as a hint for developers?

Oh, dear... Since yesterday I already came up with a 3 relatively easy ideas for format specifiers! :laughing: I will work those through during this weekend and add to this PR.

Okarss:
Did some quick tests for the new ideas and those turned out pretty badly and increased the binary size. And, indeed, as you say - the logic of all of those printf options is pretty nasty. Therefore all of the other ideas will require a significant effort if they turn out to be useful at all. So for now no more quick pull requests from me. Let's enjoy a nice release!

P.S. Don't forget the #260 .

nanoprintf.h
```diff
@@ -241,27 +242,24 @@ typedef enum {
...
 #endif
-
-  npf_format_spec_length_modifier_t length_modifier;
-  npf_format_spec_conversion_t conv_spec;
-  char case_adjust;
+  char prepend;          // ' ' or '+'
+  char alt_form;         // '#'
+  char case_adjust;      // 'a' - 'A'
+  uint8_t length_modifier;
+  uint8_t conv_spec;
```
charlesnicholson, line 262: Are there any wins from these becoming fast types?

charlesnicholson (approved):

nanoprintf.h
charlesnicholson: Well, the bottom two, I guess. The top 3 are used exclusively in addition / comparison with other char-sized types; I remember that was a win because it saved some mask instructions.

Okarss: I tried this:
```
typedef struct npf_format_spec {
#if NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS == 1
  int field_width;
  uint_fast8_t field_width_opt;
  char left_justified;   // '-'
  char leading_zero_pad; // '0'
#endif
#if NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS == 1
  int prec;
  uint_fast8_t prec_opt;
#endif
  uint_fast8_t length_modifier;
  char prepend;          // ' ' or '+'
  char alt_form;         // '#'
  char case_adjust;      // 'a' - 'A'
  uint8_t conv_spec;
} npf_format_spec_t;
```
Saves 16 B for CM7 and 2 B for RL78 binary size. No change for CM0, CM3, PIC32, Xtensa, Quark, MSP430, AVR. Increases the struct/stack size by 8 B for Cortex and probably other 32-bit CPUs. In addition to that, changing the type of the `conv_spec` element to fast type ends up mostly increasing the size. Go figure...

Okarss:
As the effects of the integer types in that structure are so strange, for now the latest commit seems to be somewhat balanced. Probably the next bigger move should be a general overhaul of the integer types for the whole project. And maybe then the effects of changing the integer types in the format specifier structure will also make more sense.

charlesnicholson:
I've noticed a very strange phenomenon where tiny changes like this seem to push gcc into bloating the function sizes up to the next 16-byte boundary or down to the previous 16-byte boundary, such that seemingly "harmless" changes will create a completely unpredictable swing of -8 / +8 bytes. I usually ignore them, because the swings change between gcc versions, which... 🤷🏻 

:)

Will merge now, thanks again for the contributions!!!

charlesnicholson (approved):

PR #271, opened 2024-08-04
Freestanding improvements (ssize_t definition and inttypes.h usage)

DeanoBurrito:
Hey again!

The issue is that when using large format specifiers nanoprintf makes use of `ssize_t`, and on non-msvc compilers that means including <sys/types.h>. This header isn't freestanding (as per the C spec), and so its existence can't be relied upon - and it often results in a compilation failure for freestanding projects. So apparently I'd patched around this in my kernel project a long time ago and forgotten, but it popped up recently when getting someone else setup with nanoprintf. So here I am opening a PR haha.

This PR is my proposed solution, but let me know what you think. I imagine if you were to merge this you'd want this documented in the readme as well - I'm also happy to do that :)

This isn't addressed here, but the other issue with compiling nanoprintf as freestanding was its use of `inttypes.h` - again something I'd patched out for my project, but not included any replacement for. Is this header still used or could it be removed?

Thanks,
Dean.

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -277,7 +277,9 @@ typedef struct npf_bufputc_ctx {
 } npf_bufputc_ctx_t;
 
 #if NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS == 1
-  #ifdef _MSC_VER
+  #ifdef NANOPRINTF_SSIZE_TYPE_OVERRIDE
+    typedef NANOPRINTF_SSIZE_TYPE_OVERRIDE ssize_t;
+  #elif defined(_MSC_VER)
     #include <BaseTsd.h>
     typedef SSIZE_T ssize_t;
   #else
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -64,7 +64,6 @@ NPF_VISIBILITY int npf_vpprintf(
 
 #include <limits.h>
 #include <stdint.h>
-#include <inttypes.h>
 
 // The conversion buffer must fit at least UINT64_MAX in octal format with the leading '0'.
 #ifndef NANOPRINTF_CONVERSION_BUFFER_SIZE
@@ -277,14 +276,10 @@ typedef struct npf_bufputc_ctx {
 } npf_bufputc_ctx_t;
 
 #if NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS == 1
-  #ifdef NANOPRINTF_SSIZE_TYPE_OVERRIDE
-    typedef NANOPRINTF_SSIZE_TYPE_OVERRIDE ssize_t;
-  #elif defined(_MSC_VER)
-    #include <BaseTsd.h>
-    typedef SSIZE_T ssize_t;
-  #else
-    #include <sys/types.h>
+  #if PTRDIFF_MAX != SIZE_MAX / 2
+    #error ptrdiff_t and size_t must be the same size
   #endif
+  typedef ptrdiff_t ssize_t;
 #endif
 
 #ifdef _MSC_VER
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -277,9 +277,9 @@ typedef struct npf_bufputc_ctx {
 
 #if NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS == 1
   //compile-time assertion
-  typedef char size_t_and_ptrdiff_t_have_same_size[(sizeof(size_t) == sizeof(ptrdiff_t)) ? 1 : -1];
+  typedef char npf_ssize_t_and_ptrdiff_t_have_same_size[(sizeof(size_t) == sizeof(ptrdiff_t)) ? 1 : -1];
 
-  typedef ptrdiff_t ssize_t;
+  typedef ptrdiff_t npf_ssize_t;
 #endif
 
 #ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


charlesnicholson:
Heya Dean-

Few things irritate me more than the vague "signed equivalent of `size_t`" language around printf, and `ssize_t` not being part of the C Standard.

How gross would it be, in your opinion, if instead of using sys/types or basetsd, nanoprintf just did a single blanket `typedef ssize_t ptrdiff_t`? It would be easy to statically assert that the size of `ptrdiff_t` is the same as the size of `size_t`, which I think they have to be given what they mean? It's important that `ssize_t` always be the same size as `size_t`, because of writeback support.

If that's not too unreasonable, it might be a nicer alternative than providing yet another npf configuration argument, though I'm not opposed to what you're pitching if there really aren't any other good solutions. Again, `ssize_t` is a very frustrating type.

(Just tossing the section 4 clause 6 text here for freestanding reference as we come back to this thread, and not to make a particular point)
Screenshot 2024-08-04 at 3 12 37 PM

DeanoBurrito:
That's a much better solution I think. Defining `ssize_t` in terms of `ptrdiff_t` does feel wrong, but I can't see any situation where it would break - and of course that could be caught with some asserts, so I think it's fine!

I took a look at some libc implementations (glibc, llvm-libc and mlibc) to see how they're defining these two types and it boils down to `int` on 32-bit platforms and `long` on 64-bit platforms, for both types. Amusingly mlibc has [a comment](https://github.com/managarm/mlibc/blob/195b55daca9cfd7b27c094d6e307afc8c668cf33/options/internal/include/bits/ssize_t.h#L5) about doing something similar to what we're talking about. 

I'll make the changes on my pr branch and push them shortly. I agree that `ssize_t` is a very frustrating type, I could only find very nebulous definitons of it.

charlesnicholson:
Sounds good to me! I'll await the PR :)

DeanoBurrito:
Let me know what you think about the check at line 279. Initially I wanted to use the `__SIZEOF_PTRDIFF_T__` (and friends) macros for gcc/clang but I couldn't find a similar mechanism for msvc - so it wasn't portable. There are the `PTRDIFF_WIDTH` and `SIZE_WIDTH` macros but they require c23. I'd welcome any suggestions around that.

charlesnicholson:
This is all pretty uncharted for me but maybe a stupid pre-c11 static assertion trick to ensure that `ptrdiff_t` and `size_t` are the same size? That would guarantee that our hacked-up `ssize_t` impostor would at least be the same size as `size_t`.

```
typedef char size_t_and_ptrdiff_t_have_same_size[(sizeof(size_t) == sizeof(ptrdiff_t)) ? 1 : -1];
```

It's kinda janky but at least it causes errors everywhere when the sizes aren't the same (I think?)

BTW I didn't realize inttypes.h wasn't required for freestanding :(

DeanoBurrito:
I didn't know that trick, thats cool! haha I've been spoiled by the features of modern c++.

Yeah `inttypes.h` isn't freestanding unfortunately. I'm not sure why, the compiler has the required info to fill it in.

nanoprintf.h
```diff
@@ -277,12 +276,10 @@ typedef struct npf_bufputc_ctx {
...
-  #ifdef _MSC_VER
-    #include <BaseTsd.h>
-    typedef SSIZE_T ssize_t;
-  #else
-    #include <sys/types.h>
-  #endif
+  //compile-time assertion
+  typedef char size_t_and_ptrdiff_t_have_same_size[(sizeof(size_t) == sizeof(ptrdiff_t)) ? 1 : -1];
+
+  typedef ptrdiff_t ssize_t;
```
charlesnicholson, line 282: I might call this `npf_ssize_t` just in case the user has included something before / after nanoprintf that defines the "real" `ssize_t`- Best if our `ssize_t` nonsense stays scoped to nanoprintf :)

nanoprintf.h
```diff
@@ -277,12 +276,10 @@ typedef struct npf_bufputc_ctx {
...
 
 #if NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS == 1
-  #ifdef _MSC_VER
-    #include <BaseTsd.h>
-    typedef SSIZE_T ssize_t;
-  #else
-    #include <sys/types.h>
-  #endif
+  //compile-time assertion
+  typedef char size_t_and_ptrdiff_t_have_same_size[(sizeof(size_t) == sizeof(ptrdiff_t)) ? 1 : -1];
```
charlesnicholson, line 280: similarly, maybe `npf_size_t_and_ptrdiff_t...` just so no poor IDE users start typing `size_t` and get this monstrosity popping up in their autocomplete :)

charlesnicholson (approved):

nanoprintf.h
DeanoBurrito: haha good point, I'll make these changes.

DeanoBurrito:
Hmm I didnt expect that, seems like the mpaland-conformance tests are relying on npf for a `ssize_t` definition.

charlesnicholson:
I just patched it, try the `freestanding` branch of the mpaland-conformance submodule- LMK how it works!

charlesnicholson:
https://github.com/charlesnicholson/nanoprintf-paland-conformance/compare/main...freestanding

DeanoBurrito:
ah nice, I just opened a PR over there, please ignore it.

charlesnicholson:
I'm pretty fast + loose over in that module; i generally just bash code into main or admin-pr-merge it immediately.

If you get a green build on that mpaland-conformance SHA, i'll merge it to main so you can retarget this PR against the new main branch of the submodule. That way this repo has no SHA on the squashed-PR main history that points to a dead SHA.

charlesnicholson:
Build is green; I just merged mpaland-conformance. Once you retarget your PR to the new main mapalnd-conformance SHA and we get another green, I'll merge! Thanks for bearing with it.

DeanoBurrito:
Great! Wont be a moment, and not a problem at all - as a user of npf I'm glad things are thorough.

charlesnicholson:
I'll update the version number and cut a release in the morning. Thanks for the patch!

DeanoBurrito:
Anytime, thanks for working with me on it!

Issue #276, opened 2025-01-22
Spurious -Wmaybe-uninitialized warning


chrisjbillington:
I'm using nanoprintf 0.5.3 and gcc 14:
```
$ gcc --version
gcc (GCC) 14.2.1 20240910
```

including nanoprintf in one of my .c files as:
```c
#define NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_BINARY_FORMAT_SPECIFIERS 1
#define NANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS 1
#define NANOPRINTF_IMPLEMENTATION
#include "nanoprintf/nanoprintf.h"
```

and compiling with (includes and source files omitted):

```
arm-none-eabi-gcc -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16 -mpure-code -Iextlibs -Wall -Wextra -ffunction-sections -fdata-sections -Wshadow -Wdouble-promotion -Wformat=2 -Wformat-truncation -Wundef -Wno-unused-parameter -fno-common  -Ofast -g -flto -Wl,--gc-sections -s -o <output>.elf <src files>
```

I get this `-Wmaybe-uninitialized` warning:

```
extlibs/nanoprintf/nanoprintf.h: In function 'npf_vpprintf.constprop':
extlibs/nanoprintf/nanoprintf.h:988:32: warning: 'fs.field_width' may be used uninitialized [-Wmaybe-uninitialized]
  988 |     field_pad = fs.field_width - cbuf_len - !!sign_c;
      |                                ^
extlibs/nanoprintf/nanoprintf.h:752:21: note: 'fs.field_width' was declared here
  752 |   npf_format_spec_t fs;
```

despite the fact that nanoprintf.h appears to contain directives to suppress this warning, which, it seems that due to aggressive optimisation, don't propagate to where they need to be to suppress the warning in all contexts.

Possible solutions:

1. initialise the struct:

```diff
--- nanoprintf.h	2025-01-23 10:25:41.108681396 +1100
[...]

PR #277, opened 2025-01-24
Initialize the npf_format_spec_t struct to prevent a spurious warning

chrisjbillington:
addresses #276 

This prevents a `-Wmaybe-uninitialized` warning on GCC, where aggressive optimizations can make the `#pragma` to suppress the warnings not work.

Remove the `#pragma` to suppress the warning as it should no longer occur.


chrisjbillington:
Let me know if you'd prefer to limit the initialisation to only happen on recent GCC versions or anything like that - or if you'd prefer to look at the `__attribute__((assume( )))` approach instead (which I also didn't previously know was a possibility - a chatbot suggested it).

chrisjbillington:
Closing in favour of #278

PR #280, opened 2025-03-25
Fix signed char conversion.

kingiler:
From [cppreference](https://en.cppreference.com/w/cpp/language/types)
> The signedness of char depends on the compiler and the target platform: the defaults for ARM and PowerPC are typically unsigned, the defaults for x86 and x64 are typically signed.

Also from [ACLE](https://arm-software.github.io/acle/main/acle.html#data-types)
> Plain char is unsigned, as specified in the ABI [AAPCS] and [AAPCS64] (7.1.1).

This fixes the failed test on `AARCH64` for `conformance.cc`
```c
TEST_CASE("conformance to system printf") {
  ...
  SUBCASE("signed int") {
    ...
    require_conform("-128", "%hhi", 128);  // print 128 instead of -128 on aarch64
    ...
  }
  ...
}
```

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -821,7 +821,7 @@ int npf_vpprintf(npf_putc pc, void *pc_ctx, char const *format, va_list args) {
           NPF_EXTRACT(NONE, int, int);
           NPF_EXTRACT(SHORT, short, int);
           NPF_EXTRACT(LONG_DOUBLE, int, int);
-          NPF_EXTRACT(CHAR, char, int);
+          NPF_EXTRACT(CHAR, signed char, int);
           NPF_EXTRACT(LONG, long, long);
 #if NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS == 1
           NPF_EXTRACT(LARGE_LONG_LONG, long long, long long);


charlesnicholson:
Ah, great catch, thanks for reporting and fixing!

I'd love to see this fail in a test before merging the patch, though. The tests today include arm64 Apple Silicon github runners, but I guess apple-clang has signed char by default because there haven't been any failures.

My thinking right now is to add a linux arm64 runner to CI in a separate PR that fails but doesn't fail the build, then rebase your PR onto main after that so we can see your PR fix the issue on the CI fleet, merge it, and then add a final PR that makes linux-arm64 fail the build.

I'm happy to do all of this, if you don't mind rebasing your PR once I have the failing arm64 linux runner up and running...

charlesnicholson:
Oof, github does provide arm64 linux runners but of course the x64 linux docker image that i created and use for linux CI doesn't work on them. Building the dockerfile on an arm image and publishing it to the container registry and pulling it down 
 will be a bigger project, so I'll merge your PR now because it looks correct to me and doesn't break any of the existing CI tests.

Thanks for the patch!

BTW feel free to add yourself to the contributors file; I'll wait a few hours before merging (or until you reply) in case you want to do that.

charlesnicholson (approved):

kingiler:
Actually, a bunch of warning and error messages on arm64 BSD, meanwhile, got stuck in loop when running test well.
I am trying to figure out what is happening. Something like `-Wexcess-padding` and `-Wunknown-warning-option` for `-Wunsafe-buffer-usage`.

charlesnicholson:
Oh, interesting- I wonder what's happening? What version of which compiler are you using? If there's a codegen path to an infinite loop I would very much like to understand that.

PR #285, opened 2025-04-18
Fix null string process

bright-pan:
char *str1 = NULL;
char *str2 = "";
char *str3 = "hello";

npf_snprintf(buf, sizeof(buf), "%s", str1); // 输出: (null)
npf_snprintf(buf, sizeof(buf), "%s", str2); // 输出: (空字符串)
npf_snprintf(buf, sizeof(buf), "%s", str3); // 输出: hello

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -806,6 +806,9 @@ int npf_vpprintf(npf_putc pc, void *pc_ctx, char const *format, va_list args) {
 
       case NPF_FMT_SPEC_CONV_STRING: {
         cbuf = va_arg(args, char *);
+        if (cbuf == NULL) {
+            cbuf = "(null)";
+        }
 #if NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS == 1
         for (char const *s = cbuf;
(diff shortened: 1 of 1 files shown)


charlesnicholson:
Thank you for the PR, but it's incorrect and I think you might need to learn more about C and nanoprintf before trying again. For example, nanoprintf prints `cbuf` in reverse, and `cbuf` points to an array that must be populated and can not be reassigned. So, for the platforms where this even compiles, if it does work, it would print `)llun(` which is not what you want. You would have discovered this if you had written a unit test, which you didn't.

You're welcome to take another attempt at this work, though- I think it could be useful for nanoprintf. But, please have it print `NULL` instead, add tests, and figure out how to do it correctly!

Good luck!

I'll leave this PR open for a few days in case you want to try fixing it here, and I'll close it if you don't respond.

charlesnicholson:
I just merged https://github.com/charlesnicholson/nanoprintf/pull/286 which prints nothing if `%s` is given NULL. If you can make it print "NULL" without increasing the binary size, I'll happily take your PR!

PR #286, opened 2025-04-19
don't print null %s values

charlesnicholson:
This PR skips null %s parameters- it doesn't dereference them and instead prints nothing.


bright-pan:
cool

PR #287, opened 2025-04-22
Buffer-based callback

stefano-zanotti-88:
This adds an option to use a buffer-based callback instead of the current single-char npf_putc.
The reduction in the number of function calls provides a substantial speedup when long strings are formatted with %s, and when the format string includes long stretches of non-% characters.
When formatting integers/floats, the buffer must be reversed before the call back, and it seems that the overall gain is quite small.
The stack occupation is slightly larger, and the code is larger.

The new kind of callback must be selected at compile time through the NANOPRINTF_USE_BUFFERED_CALLBACK option, which defaults to 0.
When it is 0, the resulting nanoprintf is identical (in behavior and footprint) to the version before the change, at least if the compiler can properly optimize the various NPF_PUTS calls that now replace inline sequences of NPF_PUTC.
Long stretches of padding characters are still output as single chars. This could be improved too, at the cost of a small local buffer to hold chunks of the padding, and at the cost of increased code size.

The new callback method does not involve copying data into an intermediate buffer. It just passes whole chunks of the argument strings or of the format string to the callback.

Note: NANOPRINTF_USE_BUFFERED_CALLBACK affects the declaration of npf_putc at the top of the nanoprintf.h file. so, I had to move the default-definition code to the "header" part, out of the implementation-only part. Maybe the checks for all-or-nothing definition of symbols, and for flag compatibility, could be moved there as well.

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -23,6 +23,23 @@
   #define NPF_PRINTF_ATTR(FORMAT_INDEX, VARGS_INDEX)
 #endif
 
+// Pick reasonable defaults if nothing's been configured.
+#if !defined(NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS) && \
+    !defined(NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS) && \
+    !defined(NANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS) && \
+    !defined(NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS) && \
+    !defined(NANOPRINTF_USE_BINARY_FORMAT_SPECIFIERS) && \
+    !defined(NANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS) && \
+    !defined(NANOPRINTF_USE_BUFFERED_CALLBACK)
+  #define NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS 1
+  #define NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS 1
+  #define NANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS 1
+  #define NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS 0
+  #define NANOPRINTF_USE_BINARY_FORMAT_SPECIFIERS 0
+  #define NANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS 0
+  #define NANOPRINTF_USE_BUFFERED_CALLBACK 0
+#endif
+
 // Public API
 
 #ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


charlesnicholson:
Hi @stefano-zanotti-88, thanks for the PR! It's interesting to me that you did this; my original design for nanoprintf was to take a `puts` callback interface exclusively, but the `putc` one was easier to make tiny and let to a simpler original implementation. I very much agree with your assessment of the optimizations- fewer callbacks == better performance. It would be very nice to revisit this interface, though, and I'm excited about the idea!

I'm on vacation for the next few days, but I want to study this and think about it. If we can figure out how to keep the footprint small, I would consider not making this interface gated by a configuration macro and instead changing nanoprintf to _only_ use this one. Migration for users is pretty trivial, I think, the `puts` callback that you supply just loops over every character, if that's what you need.

So, to my mind, the next steps are:
* I study this and think about it a bit (maybe a few days, I apologize, vacation)
* We enhance the PR to integrate the new interface into the tests- it's very important that all new npf code gets tested, especially code as fundamental as this. This will be a bit of work, I suspect, but I'm happy to help with it.
* We look at the compiled size and measure the cost of this approach.
* We make a final decision about the interface, consider doing a minor version bump, etc.

I don't know how much work you're willing to sign up for on this, but I think it's a very interesting start!

charlesnicholson:
Additionally I'd love to quantify the performance improvements- I agree that the inner loop could be offloaded to a `puts` and that's almost certainly better than the `putc` callback but I'd like to know by how much! What architecture were you testing on etc?

stefano-zanotti-88:
I'm not sure about removing the single-char putc. It offers doubtless advantages in terms of code size, so someone might prefer that to a more performant version, in constrained environments. I'd keep them both, since they're already there.

Neither I'm sure about all-or-nothing parameterization. I've added the new NANOPRINTF_USE_BUFFERED_CALLBACK option to the check, so now everyone must define it to 0 to be able to compile the library again. Maybe it would be easier on the users to just have independent parameters, that can each be defined or not, and all the non-defined ones get their default?
This would also make all the tests pass again, as a first step on which to build by adding cases for the new interface.

About the tests, I'd need guidance on how to go about that, mostly what/how to run.
Maybe it would be better to get your preliminary opinion about the code size and speed this new option achieves, before actually adapting all the tests, to avoid a bunch of work if it all turns out not to be an acceptable tradeoff for you.

I'm using nanoprintf on Cortex-M0+.
I don't have thorough benchmarking for the pre/post-change library. Do you have something of the sort, that I could run after the necessary adaptations?
The performance depends a lot on the format string and arguments, of course. I guess something very compact like printf("%c%c", a, b) is slower with the new interface. However, for long format/argument strings, the gain is large. Very approximately, from a couple % points for few-digits int/float, to 20-50% for strings with tens of characters.
The stack usage grew by 8 bytes if I'm not mistaken. The code size growth is harder to determine.

Also note that the new code still calls npf_parse_format_spec() and branches on its return value, but I'm not sure whether that's still needed. Since the added cur/end loop consumes all non-% characters, the parsing should always either succeed, or fail because of a format error, so it's possible that something can be simplified there.
What is NPF supposed to do on invalid modifiers (or modifiers that have been disabled by compile-time options)? Print them out verbatim (including/excluding the % character), or skip them altogether, or undefined behavior, or …?
The 'continue' statement in that loop might also be redundant, but I didn't investigate that.

PR #288, opened 2025-05-01
option to disable the '#' flag

stefano-zanotti-88:
'#' has limited use ("0x/0b" prefix, single "0" for octal, "." always present for floats).
Removing it --ie still parsing it, but ignoring it-- will save some code for a feature that might not be needed at all in some code bases.

If '#' is disabled, the 'alt_form' field in npf_format_spec_t is conditionally-removed.
Because of this possibility, it is be beneficial to reorder the fields of npf_format_spec_t
For instance, on systems with 2-byte integers, this reordering can reduce the struct size from 13+1 bytes to 12 bytes, if the library is configured as:
NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS    1
NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS      1
NANOPRINTF_USE_ALT_FORM_MODIFIER                0

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -79,13 +79,15 @@ NPF_VISIBILITY int npf_vpprintf(
     !defined(NANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS) && \
     !defined(NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS) && \
     !defined(NANOPRINTF_USE_BINARY_FORMAT_SPECIFIERS) && \
-    !defined(NANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS)
+    !defined(NANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS) && \
+    !defined(NANOPRINTF_USE_ALT_FORM_MODIFIER)
   #define NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS 1
   #define NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS 1
   #define NANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS 1
   #define NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS 0
   #define NANOPRINTF_USE_BINARY_FORMAT_SPECIFIERS 0
   #define NANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS 0
+  #define NANOPRINTF_USE_ALT_FORM_MODIFIER 1
 #endif
 
 // If anything's been configured, everything must be configured.
@@ -246,19 +248,23 @@ enum {
 typedef struct npf_format_spec {
 #if NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS == 1
   int field_width;
-  uint8_t field_width_opt;
-  char left_justified;   // '-'
-  char leading_zero_pad; // '0'
 #endif
 #if NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS == 1
   int prec;
   uint8_t prec_opt;
 #endif
   char prepend;          // ' ' or '+'
+#if NANOPRINTF_USE_ALT_FORM_MODIFIER == 1
   char alt_form;         // '#'
+#endif
   char case_adjust;      // 'a' - 'A'
   uint8_t length_modifier;
   uint8_t conv_spec;
+#if NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS == 1
+  uint8_t field_width_opt;
+  char left_justified;   // '-'
+  char leading_zero_pad; // '0'
+#endif
 } npf_format_spec_t;
 
 #if NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS == 0
@@ -295,7 +301,9 @@ static int npf_parse_format_spec(char const *format, npf_format_spec_t *out_spec
 #endif
   out_spec->case_adjust = 'a' - 'A'; // lowercase
   out_spec->prepend = 0;
+#if NANOPRINTF_USE_ALT_FORM_MODIFIER
   out_spec->alt_form = 0;
+#endif
 
   while (*++cur) { // cur points at the leading '%' character
     switch (*cur) { // Optional flags
@@ -305,7 +313,9 @@ static int npf_parse_format_spec(char const *format, npf_format_spec_t *out_spec
 #endif
       case '+': out_spec->prepend = '+'; continue;
       case ' ': if (out_spec->prepend == 0) { out_spec->prepend = ' '; } continue;
+#if NANOPRINTF_USE_ALT_FORM_MODIFIER
       case '#': out_spec->alt_form = '#'; continue;
+#endif
       default: break;
     }
     break;
@@ -556,7 +566,11 @@ static int npf_ftoa_rev(char *buf, npf_format_spec_t const *spec, double f) {
 
   uint_fast8_t carry; carry = 0;
   npf_ftoa_dec_t end, dec; dec = (npf_ftoa_dec_t)spec->prec;
-  if (dec || spec->alt_form) {
+  if (dec
+#if NANOPRINTF_USE_ALT_FORM_MODIFIER
+    || spec->alt_form
+#endif
+    ) {
     buf[dec++] = '.';
   }
 
@@ -879,9 +893,11 @@ int npf_vpprintf(npf_putc pc, void *pc_ctx, char const *format, va_list args) {
 #endif
         if (!val && (fs.prec_opt != NPF_FMT_SPEC_OPT_NONE) && !fs.prec) {
           // Zero value and explicitly-requested zero precision means "print nothing".
+#if NANOPRINTF_USE_ALT_FORM_MODIFIER
           if ((fs.conv_spec == NPF_FMT_SPEC_CONV_OCTAL) && fs.alt_form) {
             fs.prec = 1; // octal special case, print a single '0'
           }
+#endif
         } else
 #endif
 #if NANOPRINTF_USE_BINARY_FORMAT_SPECIFIERS == 1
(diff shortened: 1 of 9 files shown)


nanoprintf.h
```diff
@@ -295,7 +301,9 @@ static int npf_parse_format_spec(char const *format, npf_format_spec_t *out_spec
 #endif
   out_spec->case_adjust = 'a' - 'A'; // lowercase
   out_spec->prepend = 0;
+#if NANOPRINTF_USE_ALT_FORM_MODIFIER
```
charlesnicholson, line 304: Please change these tests to `#if NANOPRINTF_USE_ALT_FORM_MODIFIER == 1` (For better or worse, the rest of the file uses the `== 1` convention so I'd prefer to preserve it, and if it does change, do a sweeping change all at once).

charlesnicholson (commented):
This looks great, thanks for the PR! It looks like the tests are failing, if you can fix those up I'll happily merge it.

(Also, please add yourself to the contributors file if you like!)

nanoprintf.h
stefano-zanotti-88: I wrote "Removing it --ie still parsing it, but ignoring it--", whereas the actual code I committed skips the parsing too.
I'm not sure if you prefer the "parse but ignore" or the "fail parsing" route.
As I've commented elsewhere, we could also consider a different option for "everything that is disabled will not be honored, but will still be parsed".

I've corrected the code as you asked.

Also: if you find it easier to just take the code, tweak it, and commit it in your own branch, without having to accept the PR --which might require slower back and forth discussions with me--, that's perfectly fine. No need for changes to the contributors list or merging from my fork.


Finally, I have problems with running all the tests as specified in the readme, so it would be great if you could help me out here.
Running the build.py script, I get an HTTP error. I can give more details if you can help.
I don't understand how to build the tests myself, so right now I'm just working with a reduced set of tests in a separate file of my own.
Even if I don't use the python build, what am I supposed to compile all the "unit*.cc", "conformance.cc", "doctest_main.cc", and "paland.cc" ? All of them together don't compile. I managed to compile some of them, but the tests fail because of some conformance checks that compare implementation-defined behavior against my local standard library, which is not guaranteed to behave like NPF.

charlesnicholson: Sure, I'm happy to take this as a proof of concept and do it via a local PR at my bench if that's easier for you. I can do all the test work as well if you'd like.

Nanoprintf uses a python script to kick off CMake to generate a ninja build that compiles the 2^N flag configuration explosion. It's more than I'd do for a less-configurable project, but if nanoprintf offers a flag then I need to test that every possible build combination works with or without that flag. So, the CMakeLists.txt file has a large and very nested loop for each flag, and emits a configuration target with that particular combination of flags.

You should just be able to clone npf and run `./b --paland -v` - the python script looks for cmake and ninja in your system and, if they're absent, downloads them to the local transient "build" directory in your repo directory. (It doesn't attempt to install them or anything rude). Then, having either found or downloaded them, it simply runs cmake to configure and then ninja to build.

If that's not working for you, it might be a bug or platform incompatibility in the python script- I only test that it works on linux, so if you're on a different platform then it may not even work. (Again, the philosophy of "if it's not harnessed under CI, it's probably broken" shows itself!)

If you have the time and gumption, could I ask you to open an issue with output so that I can take a look? Generally most clients simply use nanoprintf as a submodule or just grab the .h file -- that's what it's there for -- so the build and test machinery has been lower priority. I'd still like it to work, though!

stefano-zanotti-88: Yes, it's fine if you work on your local PR.
I'll still try to run the tests, as that can speed things up.

I'm having trouble with the non-existent:
https://cmake.org/files/v3.22/cmake-3.22.2-win64-x86_64.tar.gz
which makes the build script fail.
I'll try and download cmake manually when I have time, and go on from there.
Do you think cmake 3.31 vs 4.0 should make any difference?

I'm on Windows, by the way. So I ignore "b" and run `python build.py`

charlesnicholson: I just landed a PR that tests windows downloading and fixed a few issues- cmake changed its download file naming scheme and extension for windows, which explains and fixes the error you saw.

So, `python build.py --download --paland -v` will force the downloading and sandboxing of the latest ninja + cmake versions, and give you as close of a local experience as possible to what the build server is doing.

If you're curious, you can go here: https://github.com/charlesnicholson/nanoprintf/actions/runs/14815979224/job/41596450464?pr=300 
and expand the "Build" tab to see what it's doing. Hopefully it will be the same on your computer? Please let me know if you try it out.

I'll start the PR integration work tomorrow, by the way, thanks again for all the PRs and very thoughtful discussion!

stefano-zanotti-88: You're welcome, and thanks for the fix for build.py.

I'm trying to run the tests locally.
I've adapted the build to consider the new option NANOPRINTF_USE_ALT_FORM_FLAG. Please have a look at that, see if it looks good to you.
Anyway, I cannot run the tests, I get this error:
```
[3/805] C:\Program_Files\MinGW\bin\c++.exe -DDOCTEST_CONFIG_SUPER_FAST_ASSERTS -DNANOPRINTF_USE_ALT_FORM_FLAG=1 -DNANOPRINTF_USE_BINARY_FORMAT_SPECIFIERS=0 -DNANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS=0 -DNANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS=0 -DNANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS=0 -DNANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS=0 -DNANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS=1  -Os -std=gnu++17 -pedantic -Wall -Wextra -Wundef -Werror -Wconversion -Wshadow -Wfloat-equal -Wsign-conversion -Wswitch-enum -Wswitch-default -MD -MT CMakeFiles/npf_conform_writeback_altform.dir/tests/conformance.cc.obj -MF CMakeFiles\npf_conform_writeback_altform.dir\tests\conformance.cc.obj.d -o CMakeFiles/npf_conform_writeback_altform.dir/tests/conformance.cc.obj -c C:/npf/tests/conformance.cc
FAILED: CMakeFiles/npf_conform_writeback_altform.dir/tests/conformance.cc.obj 
C:\Program_Files\MinGW\bin\c++.exe -DDOCTEST_CONFIG_SUPER_FAST_ASSERTS -DNANOPRINTF_USE_ALT_FORM_FLAG=1 -DNANOPRINTF_USE_BINARY_FORMAT_SPECIFIERS=0 -DNANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS=0 -DNANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS=0 -DNANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS=0 -DNANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS=0 -DNANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS=1  -Os -std=gnu++17 -pedantic -Wall -Wextra -Wundef -Werror -Wconversion -Wshadow -Wfloat-equal -Wsign-conversion -Wswitch-enum -Wswitch-default -MD -MT CMakeFiles/npf_conform_writeback_altform.dir/tests/conformance.cc.obj -MF CMakeFiles\npf_conform_writeback_altform.dir\tests\conformance.cc.obj.d -o CMakeFiles/npf_conform_writeback_altform.dir/tests/conformance.cc.obj -c C:/npf/tests/conformance.cc
C:/npf/tests/conformance.cc: In function 'void DOCTEST_ANON_FUNC_2()':
C:/npf/tests/conformance.cc:404:46: error: too many arguments for format [-Werror=format-extra-args]
  404 |     npf_pprintf(+[](int, void*) {}, nullptr, "1234567%hhn", &writeback);
      |                                              ^~~~~~~~~~~~~
cc1plus.exe: all warnings being treated as errors
```
I'm using GCC:
gcc (x86_64-win32-seh-rev0, Built by MinGW-Builds project) 15.1.0

If I'm not mistaken, this MinGW build includes an MSVC runtime, so maybe it lacks support for "hh" modifiers?
Do you think I should try with another MinGW, or with clang?

Finally, I'm still trying to wrap my head around how PRs work, so for now I just cherry-picked the build.py fix into this branch. Let me know what would be a better way to go about such things. Could I have rebased this branch onto your fix? Would this PR still be valid in that case, since the original branch would be gone, and the rebased one be effectively a new one with the same name?

stefano-zanotti-88: I don't understand why the new flag NANOPRINTF_USE_ALT_FORM_FLAG is not being passed as a compilation option. I've added it in CMakeLists.txt, but apparently that wasn't sufficient.

charlesnicholson: I've got a few hours today that I'm going to dedicate to landing your PRs; I'll dig into all of this stuff.

Re: the windows build failing, it looks like it's using gcc through mingw- I updated the build instructions to indicate that the windows build today is only supported for MSVC. I don't think many users actually use nanoprintf outside of embedded contexts, but sometimes it's handy for host-side firmware library builds for unit / functional tests etc. Anyway, I chose MSVC because it has very different opinions on warnings, whereas just pushing it through gcc on yet another platform didn't seem helpful. (The macOS build looks like clang, but it's actually apple's custom clang that they ship with Xcode and it also has different warnings / errors).

So, I think the last step to getting you up and running is for you to launch the nanoprintf build from inside of a visual studio developer command prompt. That way CMake will find cl.exe and assume that it's an MSVC build. I'm guessing you're running from git bash, so cmake is finding the mingw gcc toolchain?

stefano-zanotti-88: I'm running `python build.py` from a cmd prompt. I have MinGW "installed" (placed somewhere reachable from the system PATH).
I just tried building from git bash, and the error is the same -- it is still finding MinGW and compiling with that.
I don't know how to run it "inside of a visual studio developer command prompt". Do you mean running `python build.py` in the console of a VS code installation, or of full Visual Studio? I'm not familiar with that, but I could try. Or I could download clang and get the build script to find that instead of MinGW's GCC.

charlesnicholson: Fascinating- sorry you're going through this; you're challenging a bunch of my assumptions about windows development :) I was in videogames + windows app development for a decade or so, and having a full Visual Studio installation was something that everybody always had for any software projects, so I'm clearly taking that for granted.

When you have Visual Studio installed for C/C++ development, it installs a shortcut that will launch powershell or the command prompt with various environment variables set up. Those update your PATH to include directories that contain programs like "cl.exe" and "link.exe" which are used for command-line compilation via MSVC. When those path entries are present, CMake will discover those tools during its host-platform-interrogation phase, and say "oh, I see, I'm doing a visual studio build." You can see CI run the script that sets the environment up for 32- or 64-bit compilation mode here: https://github.com/charlesnicholson/nanoprintf/blob/main/.github/workflows/presubmit.yml#L196

There's no particular reason that nanoprintf shouldn't work with mingw-based setups; I just haven't done the work to set it up because it's not how I develop when I'm on windows!

stefano-zanotti-88: I had a VS Code lying around from 2020, with the ms-vscode.cpptools extension installed.
I tried building from its terminal but it still finds MinGW. I hoped it would come with the necessary tools you described, but apparently that's only with full VS.
I'll see if I can get it, though it was already really heavyweight the last time I installed it years ago.

charlesnicholson: I've never tried it through VS Code; I always do the (yes, really heavyweight) full VS installation.

charlesnicholson:
Implemented in https://github.com/charlesnicholson/nanoprintf/pull/304, mostly just copy-paste. thanks for the PR!

PR #290, opened 2025-05-02
Fixed wrong sign for negative zero

stefano-zanotti-88:
The standard says, in a footnote:
```
The results of all floating conversions of a negative zero, and of negative values that round to zero, include a minus sign.
```

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -515,7 +515,7 @@ enum {
   NPF_DOUBLE_EXP_BIAS = DBL_MAX_EXP - 1,
   NPF_DOUBLE_MAN_BITS = DBL_MANT_DIG - 1,
   NPF_DOUBLE_BIN_BITS = sizeof(npf_double_bin_t) * CHAR_BIT,
-  NPF_DOUBLE_SIGN_POS = sizeof(double) * CHAR_BIT - 1, // this is true for f32 and f64; if we were paranoid, we should calculate it as (NPF_DOUBLE_MAN_BITS + NPF_DOUBLE_EXP_BITS), where NPF_DOUBLE_EXP_BITS is ceil(log2(DBL_MAX_EXP)), hard to get as a compile-time constant unless we check against all the possible powers-of-2
+  NPF_DOUBLE_SIGN_POS = sizeof(double) * CHAR_BIT - 1, // this is true for f32 and f64; if we were paranoid, we should calculate it as (NPF_DOUBLE_MAN_BITS + NPF_DOUBLE_EXP_BITS), where NPF_DOUBLE_EXP_BITS is ceil(log2(DBL_MAX_EXP-DBL_MIN_EXP)), hard to get as a compile-time constant unless we check against all the possible powers-of-2
   NPF_FTOA_MAN_BITS   = sizeof(npf_ftoa_man_t) * CHAR_BIT,
   NPF_FTOA_SHIFT_BITS =
     ((NPF_FTOA_MAN_BITS < DBL_MANT_DIG) ? NPF_FTOA_MAN_BITS : DBL_MANT_DIG) - 1
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -515,6 +515,7 @@ enum {
   NPF_DOUBLE_EXP_BIAS = DBL_MAX_EXP - 1,
   NPF_DOUBLE_MAN_BITS = DBL_MANT_DIG - 1,
   NPF_DOUBLE_BIN_BITS = sizeof(npf_double_bin_t) * CHAR_BIT,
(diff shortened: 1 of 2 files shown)


nanoprintf.h
```diff
@@ -680,6 +681,20 @@ static int npf_ftoa_rev(char *buf, npf_format_spec_t const *spec, double f) {
...
+  // alternatively to this function, we could use the standard signbit() from math.h
+
+  npf_double_bin_t bin; { // Union-cast is UB pre-C11, compiler optimizes byte-copy loop.
+    char const *src = (char const *)&f;
+    char *dst = (char *)&bin;
+    for (uint_fast8_t i = 0; i < sizeof(f); ++i) { dst[i] = src[i]; }
+  }
+  return (bin >> NPF_DOUBLE_SIGN_POS) & 0x1;
+}
+
```
charlesnicholson, line 697: I'd like to explore a helper function like:
```
static inline double npf_double_to_int_rep(double f) {
 npf_double_bin_t bin; { // Union-cast is UB in C++ and pre-C11, compiler optimizes byte-copy loop.
    char const *src = (char const *)&f;
    char *dst = (char *)&bin;
    for (uint_fast8_t i = 0; i < sizeof(f); ++i) { dst[i] = src[i]; }
  }
  return bin;
}
```
Then this helper function could be used at the beginning of `npf_ftoa_rev`, as well as in-line later.

nanoprintf.h
```diff
@@ -944,7 +959,7 @@ int npf_vpprintf(npf_putc pc, void *pc_ctx, char const *format, va_list args) {
           val = va_arg(args, double);
         }
 
-        sign_c = (val < 0.) ? '-' : fs.prepend;
+        sign_c = npf_signbit(val) ? '-' : fs.prepend;
```
charlesnicholson, line 962: Then this would become something like
```
sign_c = (npf_double_to_int_rep(val) >> NPF_DOUBLE_SIGN_POS) ? '-' : fs.prepend;
```

nanoprintf.h
```diff
@@ -515,6 +515,7 @@ enum {
   NPF_DOUBLE_EXP_BIAS = DBL_MAX_EXP - 1,
   NPF_DOUBLE_MAN_BITS = DBL_MANT_DIG - 1,
   NPF_DOUBLE_BIN_BITS = sizeof(npf_double_bin_t) * CHAR_BIT,
+  NPF_DOUBLE_SIGN_POS = sizeof(double) * CHAR_BIT - 1, // this is true for f32 and f64; if we were paranoid, we should calculate it as (NPF_DOUBLE_MAN_BITS + NPF_DOUBLE_EXP_BITS), where NPF_DOUBLE_EXP_BITS is ceil(log2(DBL_MAX_EXP-DBL_MIN_EXP)), hard to get as a compile-time constant unless we check against all the possible powers-of-2
```
charlesnicholson, line 518: I don't quite understand your comment yet; are there situations where `sizeof(double) * CHAR_BIT - 1` won't work? I understand that the size-in-bytes of float and double aren't hard-coded but specified in terms of precisions, to allow for exciting architectures, but are there environments where this will be incorrect?

Could I ask you for a little more clarity about the "why" in this comment, if you feel that this comment should be left in?

charlesnicholson (commented):
Thanks for the PR! I left some comments in-line, but the idea of extracting `npf_double_to_int_rep` only makes sense if we can establish that the compiler inlines it. I'd be in favor of a best-effort `NPF_FORCE_INLINE` macro that's the opposite of the `NPF_NOINLINE` macro; I certainly don't want to spill a uint64_t to the stack if the function can't be inlined.

nanoprintf.h
stefano-zanotti-88: C doesn't require very much regarding `double`. It could even be a non-base-2 representation.
The sign bit is not required to be the MSb. To be true, NPF obviously relies on double being an IEEE 754 float, otherwise even the placement of exponent and mantissa are not necessarily the ones we assume. Nor is the representation of nans and infinities.
I'm not sure about possible future IEEE 754 formats, but I'd guess the sign bit is going to stay there.
What is said in the comment, after all, still relies on IEEE 754, it just explicitly calculates how many bits are needed to skip over the exponent field.

Let's just remove that comment.

nanoprintf.h
stefano-zanotti-88: I'm in favor of `npf_double_to_int_rep` too.

This is what I'd do for the inline attribute:
```
#if defined(__clang__) || defined(__GNUC__) || defined(__GNUG__)
  #define NPF_FORCE_INLINE  __attribute__((always_inline))
#elif defined(_MSC_VER)
  #define NPF_FORCE_INLINE  __forceinline 
#else
  #define NPF_FORCE_INLINE 
#endif
```

I don't know how you'd go about checking whether the function is actually inlined, across different platforms.
Such a function seems very easily inlineable, though, if the compiler can actually remove the byte copy (which is something we're already relying on).

Alternatively, this ugly thing could be used:
```
#define NPF_DBL_TO_INT_REP(/* npf_double_bin_t */ OUT, /* double */ DBL)    do { \
    npf_double_bin_t bin; { /* Union-cast is UB pre-C11, compiler optimizes byte-copy loop. */ \
      char const *src = (char const *)&DBL; \
      char *dst = (char *)&bin; \
      for (uint_fast8_t i = 0; i < sizeof(f); ++i) { dst[i] = src[i]; } \
    } \
    OUT = bin; \
  } while(0)
```

PR #292, opened 2025-05-02
Less surprising UB for width/precision associated to %c %s %%

stefano-zanotti-88:
Before this fix, we have (NPF vs other implementations):
```
printf("%10c", 'a');    // "         a" vs "         a"
printf("%010c", 'a');   // "a"          vs "000000000a"
printf("%10s" , "abc"); // "       abc" vs "       abc"
printf("%010s", "abc"); // "       abc" vs "0000000abc"
printf("%010%");        // "%"          vs "%"
printf("%10%");         // "         %" vs "%"
printf("%.2c" , 'a');   // "0a" vs "a"
printf("%.2s" , "abc"); // "ab" vs "ab"
printf("%.2%");         // "0%" vs "%"
```
Afterwards:
```
printf("%10c", 'a');    // "         a"
printf("%010c", 'a');   // "000000000a"
printf("%10s" , "abc"); // "       abc"
printf("%010s", "abc"); // "0000000abc"
printf("%010%");        // "000000000%"
printf("%10%");         // "         %"
printf("%.2c" , 'a');   // "a"
printf("%.2s" , "abc"); // "ab"
printf("%.2%");         // "%"
```

The standard describes the width field in general, for any conversion specifier.
It makes no special provisions for 'c'. Compare with 'n', for which it is explicitly stated:
```
If the conversion specification includes any flags, a field width, or a precision, the behavior is undefined.
```
and also for '%':
```
The complete conversion specification shall be %%
```
So, I would say that 'c' should respect the field width just like 's'.

The problem for NPF originates from the '0' flag, which is illegal for 'c', so we have UB, and NPF is technically conforming. See the dedicated code around line 960.
However, padding with '0' is less surprising than just ignoring the width altogether. It's also more consistent across 'c' and 's' and '%'.
This also removes some checks, resulting in a smaller code footprint, while remaining standard-conforming.

Notice that ignoring '0' was done around line 393 for 's', which caused the different treatment of 'c' and 's' around line 960.

Similarly, I've removed some checks for '%', which make it more consistent (still UB, hence conforming) and smaller.
Finally, I've and added a couple prec=0 assignments to avoid spurious '0' padding when we ignore the precision, though this means more code.

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -380,19 +380,18 @@ static int npf_parse_format_spec(char const *format, npf_format_spec_t *out_spec
     case '%': out_spec->conv_spec = NPF_FMT_SPEC_CONV_PERCENT;
 #if NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS == 1
       out_spec->prec_opt = NPF_FMT_SPEC_OPT_NONE;
+      out_spec->prec = 0;
 #endif
       break;
 
     case 'c': out_spec->conv_spec = NPF_FMT_SPEC_CONV_CHAR;
 #if NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS == 1
       out_spec->prec_opt = NPF_FMT_SPEC_OPT_NONE;
+      out_spec->prec = 0;
 #endif
       break;
 
     case 's': out_spec->conv_spec = NPF_FMT_SPEC_CONV_STRING;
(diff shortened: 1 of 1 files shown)


charlesnicholson:
Thanks for all of these PRs! I'm going to dive in this weekend and start working through them :)

Also if you'd like, please consider a PR that adds your name to the contributors list.

charlesnicholson:
Implemented in https://github.com/charlesnicholson/nanoprintf/pull/308, thanks for yet another PR!

PR #293, opened 2025-05-03
Removed default '#' from %p

stefano-zanotti-88:
%p is implementation-defined, so the current implementation is fine.
However, it introduces a default '#'. This is present in other implementations as well, though not all of them.
%p respects width and precision specifiers, as well as flags.
So, it seems inconsistent that it should not respect the presence/absence of '#'.

Before vs after this change:
```
printf("%p"     , NULL);     // "0x0"          vs "0"
printf("%#p"    , NULL);     // "0x0"          vs "0"
printf("%10p"   , NULL);     // "       0x0"   vs "         0"
printf("%#10p"  , NULL);     // "       0x0"   vs "         0"
printf("%010p"  , NULL);     // "0x00000000"   vs "0000000000"
printf("%#010p" , NULL);     // "0x00000000"   vs "0000000000"
printf("%.10p"  , NULL);     // "0x0000000000" vs "0000000000"
printf("%#.10p" , NULL);     // "0x0000000000" vs "0000000000"
printf("%0.10p" , NULL);     // "0x0000000000" vs "0000000000"
printf("%#0.10p", NULL);     // "0x0000000000" vs "0000000000"
printf("%p"     , (void*)1); // "0x1"          vs "1"
printf("%#p"    , (void*)1); // "0x1"          vs "0x1"
printf("%10p"   , (void*)1); // "       0x1"   vs "         1"
printf("%#10p"  , (void*)1); // "       0x1"   vs "       0x1"
printf("%010p"  , (void*)1); // "0x00000001"   vs "0000000001"
printf("%#010p" , (void*)1); // "0x00000001"   vs "0x00000001"
printf("%.10p"  , (void*)1); // "0x0000000001" vs "0000000001"
printf("%#.10p" , (void*)1); // "0x0000000001" vs "0x0000000001"
printf("%0.10p" , (void*)1); // "0x0000000001" vs "0000000001"
printf("%#0.10p", (void*)1); // "0x0000000001" vs "0x0000000001"
```

You could argue that with this change, %p is mostly useless, as it is equivalent to %x with appropriate parameters; however, that's mostly true in any case, and the major benefit of %p is that you don't need a cast to the appropriate integer type (like casting to uintptr_t and using %tx, or casting to (unsigned long) and using %lx, or something else, depending on your platform).

Indeed, it could be easier just to change line 914 from:
```
need_0x = 'x';
```
to
```
if (fs.alt_form) { need_0x = 'x'; }
```
Note that this, just like the current behavior, always adds "0x", whereas the "%x" equivalent proposed in this change omits it when the number is 0 (a very weird corner case required by the standard).

I'd say it could be more logical, and also result in smaller code, to apply the one-line change instead.
Also, this would allow for an easier removal of %p support, which would reduce the code footprint. I'm sure that not everybody has a need for %p.

Also note: if '#' is made optional here, and you accept #288 (making '#'-support optional for NPF), then further work would be needed to completely remove the 'need_0x' logic when '#'-support is disabled.

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -855,22 +855,27 @@ int npf_vpprintf(npf_putc pc, void *pc_ctx, char const *format, va_list args) {
 #endif
       case NPF_FMT_SPEC_CONV_OCTAL:
       case NPF_FMT_SPEC_CONV_HEX_INT:
-      case NPF_FMT_SPEC_CONV_UNSIGNED_INT: {
+      case NPF_FMT_SPEC_CONV_UNSIGNED_INT:
+      case NPF_FMT_SPEC_CONV_POINTER: {
         npf_uint_t val = 0;
 
-        switch (fs.length_modifier) {
-          NPF_EXTRACT(NONE, unsigned, unsigned);
-          NPF_EXTRACT(SHORT, unsigned short, unsigned);
-          NPF_EXTRACT(LONG_DOUBLE, unsigned, unsigned);
-          NPF_EXTRACT(CHAR, unsigned char, unsigned);
-          NPF_EXTRACT(LONG, unsigned long, unsigned long);
+        if (fs.conv_spec == NPF_FMT_SPEC_CONV_POINTER) {
+          val = (npf_uint_t)(uintptr_t)va_arg(args, void *);
+        } else {
+          switch (fs.length_modifier) {
+            NPF_EXTRACT(NONE, unsigned, unsigned);
+            NPF_EXTRACT(SHORT, unsigned short, unsigned);
+            NPF_EXTRACT(LONG_DOUBLE, unsigned, unsigned);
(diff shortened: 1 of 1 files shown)


charlesnicholson:
This is very insightful and IMO a nice improvement to the code size- if users really want %p to add the 0x, they can add the # themselves.

I'm happy to merge this but the tests need some updating. I also think your other proposal in #288 is good so I think there's now an ordering issue. What order would you like to proceed in for these PRs?

stefano-zanotti-88:
I'd go with #288 first.
I'll fix the tests.
Let me know which route you like best: the single-line change, or the actual content of the PR, or something else.

charlesnicholson:
Implemented as https://github.com/charlesnicholson/nanoprintf/pull/305 - thanks for the PR!

PR #294, opened 2025-05-03
Made 'h' and 'hh' optional

stefano-zanotti-88:
Integer types passed through variadic arguments are promoted to int, if they are of a narrower type.
Specifiers 'h' and 'hh' let you specify the original type (ie short/char), so that the library can retrieve the value using the proper type.
The C23 standard (here the latest open draft: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf), around §7.23.6.1, says that you should retrieve that value, cast it to the appropriate type, and then print it.
NPF does that correctly.

However, this is only useful in weird corner cases, that is, when the caller passes a value larger than the type that was specified.
For instance (assuming 16-bit shorts and 32-bit ints):
printf("%hx", 1u); // no problem, will print "1"
printf("%hx", (unsigned short)0xAFFFF); // no problem, will print "ffff"; the value that is passed to the function is (int)0x0000FFFF
printf("%hx", 0xAFFFF); // problem: must print "ffff", not "affff"; the value that is passed to the function is (int)0x000AFFFF
So, the cast is indeed necessary in the general case. However, I can't think of a reason why the caller should ever trigger such corner cases; and it can always prevent the problem by casting the value appropriately.
So, the library can be smaller by just ignoring the 'h' and 'hh' modifiers. This will cause the library to take the code path for NPF_EXTRACT(int, int), which is fine for normal cases.
You could also completely remove the parsing of 'h' and 'hh'. This can cause more problems, since the caller can no longer use such types then (and such types might also arise "non-transparently" from using `PRI***16` and `PRI***8`).
However, keeping the parsing means that we just remove a couple of casts in the extract/writeback code paths.

Before vs after (on a platform with 1/2/4 char/short/int):
```
printf("%hhu", (unsigned char)1u);       // "1"     vs "1"
printf("%hhu", 1u);                      // "1"     vs "1"
printf("%hhu", (unsigned char)UINT_MAX); // "255"   vs "255"
printf("%hhu", UINT_MAX);                // "255"   vs "4294967295"
printf("%hu", (unsigned short)1u);       // "1"     vs "1"
printf("%hu", 1u);                       // "1"     vs "1"
printf("%hu", (unsigned short)UINT_MAX); // "65535" vs "65535"
printf("%hu", UINT_MAX);                 // "65535" vs "4294967295"
printf("%hhi", (signed char)1);          // "1"     vs "1"
printf("%hhi", 1);                       // "1"     vs "1"
printf("%hhi", (signed char)INT_MIN);    // "0"     vs "0"
printf("%hhi", INT_MIN);                 // "0"     vs "-2147483648"
printf("%hhi", (signed char)INT_MAX);    // "-1"    vs "-1"
printf("%hhi", INT_MAX);                 // "-1"    vs "2147483647"
printf("%hi", (short)1);                 // "1"     vs "1"
printf("%hi", 1);                        // "1"     vs "1"
printf("%hi", (short)INT_MIN);           // "0"     vs "0"
printf("%hi", INT_MIN);                  // "0"     vs "-2147483648"
printf("%hi", (short)INT_MAX);           // "-1"    vs "-1"
printf("%hi", INT_MAX);                  // "-1"    vs "2147483647"
```

In this PR, I've added 3 levels of support (so that we can investigate the tradeoffs, and decide if any of them is worthwhile): no support (not even for %n), support for parsing (but not proper casting), full support.
They are not exposed to the user yet

Related to fetching variadic arguments: why is there a case for NPF_EXTRACT(LONG_DOUBLE, int, int) on lines 823 and 864, and one for NPF_WRITEBACK(LONG_DOUBLE, double) on line 923?
The 'L' specifier is already handled for float specifiers. 'L' with an integer specifier is not allowed. Is this code just for robustness in case of a malformed format string? You could just as well let the parsing fail in this case. I've removed those lines too, in this PR.

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -88,6 +88,8 @@ NPF_VISIBILITY int npf_vpprintf(
   #define NANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS 0
 #endif
 
+#define NANOPRINTF_USE_SMALL_FORMAT_SPECIFIERS    1 // 0: no;  1: parsing only;  2: also proper casting when fetching arguments
+
 // If anything's been configured, everything must be configured.
 #ifndef NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS
   #error NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS must be #defined to 0 or 1
@@ -207,9 +209,10 @@ enum {
 
 enum {
   NPF_FMT_SPEC_LEN_MOD_NONE,
+#if NANOPRINTF_USE_SMALL_FORMAT_SPECIFIERS
   NPF_FMT_SPEC_LEN_MOD_SHORT,       // 'h'
-  NPF_FMT_SPEC_LEN_MOD_LONG_DOUBLE, // 'L'
   NPF_FMT_SPEC_LEN_MOD_CHAR,        // 'hh'
+#endif
   NPF_FMT_SPEC_LEN_MOD_LONG,        // 'l'
 #if NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS == 1
   NPF_FMT_SPEC_LEN_MOD_LARGE_LONG_LONG, // 'll'
@@ -217,6 +220,7 @@ enum {
   NPF_FMT_SPEC_LEN_MOD_LARGE_SIZET,     // 'z'
   NPF_FMT_SPEC_LEN_MOD_LARGE_PTRDIFFT,  // 't'
 #endif
+  NPF_FMT_SPEC_LEN_MOD_LONG_DOUBLE, // 'L'
 };
 
 enum {
@@ -349,6 +353,7 @@ static int npf_parse_format_spec(char const *format, npf_format_spec_t *out_spec
   uint_fast8_t tmp_conv = NPF_FMT_SPEC_CONV_NONE;
   out_spec->length_modifier = NPF_FMT_SPEC_LEN_MOD_NONE;
   switch (*cur++) { // Length modifier
+#if NANOPRINTF_USE_SMALL_FORMAT_SPECIFIERS
     case 'h':
       out_spec->length_modifier = NPF_FMT_SPEC_LEN_MOD_SHORT;
       if (*cur == 'h') {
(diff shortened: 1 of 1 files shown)


charlesnicholson:
> Related to fetching variadic arguments: why is there a case for NPF_EXTRACT(LONG_DOUBLE, int, int) on lines 823 and 864, and one for NPF_WRITEBACK(LONG_DOUBLE, double) on line 923?
The 'L' specifier is already handled for float specifiers. 'L' with an integer specifier is not allowed. Is this code just for robustness in case of a malformed format string? You could just as well let the parsing fail in this case. I've removed those lines too, in this PR.

Nope, it was just a dumb oversight- I have no idea why I added that, it's very very old code from the original bringup, many years ago :)

@ZERICO2005 just noticed it as well over in this https://github.com/charlesnicholson/nanoprintf/issues/311 - I just fixed it in a tiny standalone PR. (Simpler to separate concerns, smaller PRs are easier than bigger ones, etc)

charlesnicholson:
I agree with what you're writing, but personally side with standard adherence and user-expectation on this one-

It's strange / fringe but I feel that `printf("%hhx", 0x12345678);` should print `78` through nanoprintf. Users can always cast and use `%x` as you pointed out to achieve the same result, but we have `h` and `hh` so that they don't need to.

I have been frustrated in the past (infrequently, I admit!) that newlib-nano didn't support `hh` at all last time I checked it; this is a dusty corner of printf; I'd rather get it right and pay the small number of extra bytes to support it.

I'd much rather pursue your other idea that we have a new flag called `NANOPRINTF_USE_SMALL_FORMAT_SPECIFIERS` that's set to 1 by default, and completely removes all support from the parser. That makes the ultra-min-spec build a bit smaller, but for less constrained use, I expect most users will leave it enabled and ignore it.

stefano-zanotti-88:
Yours is a perfectly reasonable stance.
It's not like we can have huge gains by departing from the standard on this one.

Anyway, I've never had any use for "h" and "hh". Why would I want to use it and pass in a value of the wrong type?
If I really want to cast, I can cast explicitly myself.
Also, small types are always promoted, so why not just avoid "hh" and "h" altogether?
I can only think of them as useful for %n, since in that case you really cannot just use the promoted type. But I've never had any use for %n (of any size) either, so it doesn't make much difference to me.
These are just thoughts about printf, not about npf.
What do you find these modifiers helpful for?

I agree with your choice about NANOPRINTF_USE_SMALL_FORMAT_SPECIFIERS.

PR #296, opened 2025-05-03
Use 'restrict' pointers

stefano-zanotti-88:
In the standard printf functions (in C, not C++), the output buffer/file and the format string are declared as 'restrict' pointers (char* restrict, const char* restrict).
This was introduced in C99, at least according to https://en.cppreference.com/w/c/io/fprintf

Restrict pointers should help the compiler optimize access to those arrays.
It could be interesting to see the code size changes across various platforms.

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -16,7 +16,7 @@
   #define NPF_VISIBILITY extern
 #endif
 
-#if !defined(NANOPRINTF_NO_RESTRICT) && (__STDC_VERSION__ + 0) >= 199901L  // restrict only exists in C, not C++
+#if !defined(NANOPRINTF_NO_RESTRICT) && defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L  // restrict only exists in C, not C++
   #define NPF_RESTRICT    restrict
 #else
   #define NPF_RESTRICT
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -16,6 +16,12 @@
   #define NPF_VISIBILITY extern
 #endif
 
+#if !defined(NANOPRINTF_NO_RESTRICT) && (__STDC_VERSION__ + 0) >= 199901L  // restrict only exists in C, not C++
+  #define NPF_RESTRICT    restrict
+#else
+  #define NPF_RESTRICT
+#endif
+
 #if defined(__clang__) || defined(__GNUC__) || defined(__GNUG__)
   #define NPF_PRINTF_ATTR(FORMAT_INDEX, VARGS_INDEX) \
     __attribute__((format(printf, FORMAT_INDEX, VARGS_INDEX)))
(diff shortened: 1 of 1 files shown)


nanoprintf.h
```diff
@@ -16,6 +16,12 @@
   #define NPF_VISIBILITY extern
 #endif
 
+#if !defined(NANOPRINTF_NO_RESTRICT) && (__STDC_VERSION__ + 0) >= 199901L  // restrict only exists in C, not C++
+  #define NPF_RESTRICT    restrict
+#else
+  #define NPF_RESTRICT
+#endif
```
charlesnicholson, line 23: This is causing the C++ compilation tests to fail because `__STDC_VERSION__` isn't defined in C++. I suspect that this test needs to become slightly more complex, and also test `!defined(__cplusplus)`.

charlesnicholson (commented):
Once this passes CI, we can look at the "size reports" job to at learn if the popular compilers can use this change to skip any spurious reloads. I'm not against this PR, but if it doesn't help anything today then it's a little speculative and means that we should either add a platform / toolchain that does have an impact or rethink it.

(nanoprintf is pretty complex code because of all of the ifdefs and macros, so I'm nervous to take performance-only PRs that don't show benefits in the current CI)

nanoprintf.h
stefano-zanotti-88: That `+ 0` is there exactly to avoid problems when __STDC_VERSION__ is not defined, either in C++, or in C89. It results in `(+0) >= 199901L` evaluating to false, which is what we want in that case, ie no restrict.
It must be because of "-Wundef". I think I've fixed it now.

And yes, this PR is just an idea, to explore if it can have any impact at all.

stefano-zanotti-88: Using -Wundef is indeed a good idea.
Just for information, these are symbols that NPF treats as defined/undefined, rather than defined to 0/1:
```
NANOPRINTF_VISIBILITY_STATIC
NANOPRINTF_IMPLEMENTATION
NPF_HAVE_BUILTIN_CLZ
NANOPRINTF_SNPRINTF_SAFE_EMPTY_STRING_ON_OVERFLOW
```
Maybe it would be good to turn all of them into always-defined 0/1?
That would break backwards compatibility though.

Just an aside:
To still move these symbols to 0/1, and retain compatibility with "defined as empty means enabled", you need quite some acrobatics:
```
// Turn an "undefined/defined" macro X into a "0/1 always-defined" macro
#if defined(X) && (0 - X - 1) == 1 && (X - 0) != -2
  // If X is defined, and is empty.
  // It works because:
  //   if empty:
  //     (0 -  - 1) == 1 && ( - 0) != -2
  //     +1 == 1 && 0 != 2
  //     true && true
  //   if non-empty (== -2):
  //     (0 - -2 - 1) == 1 && (-2 - 0) != -2
  //     1 == 1 && -2 != -2
  //     true && false
  //   if non-empty (else):
  //     (0 - <else> - 1) == 1 && (<else> - 0) != -2
  //     false && <irrelevant>
  #undef X
  #define X    <default "enabled" value, possibly 1>
#elif !defined(X)
  // If X is not defined
  #define X    <default "disable" value>  // or just raise an error
#else
  // If X is defined as non-empty
  // just keep it
#endif
```
This:
* fails with a compilation error for (most) non-integer values
* retains the user-provided value for any integer value
* turns a "defined-as-empty" symbol into whatever you want
* provides a default value if X was never defined, or treats it as an error

Well, technically this misdetects symbols like the following as being "defined as empty":
```
#define X    0,2
#define X    0|2
#define X    0==1
```
but I don't know how one could do any better.
The latter one in particular could occur if the user defines it using some condition like
```
#define X     MY_APP_NEEDS_FLOATS==1
```
And we cannot put X inside parenthesis in the checks above, since an empty X would produce "0 - () - 1" which fails to parse.

charlesnicholson: Requiring all user-supplied macros to be explicitly defined to 0 or 1 would be a very welcome standalone PR! I'd prefer not to combine it with the restrict implementation though.

I want to think about the macro math approach; I get what you're doing but it's very complex and I'd prefer not to shoulder the readability burden of it. Something along these lines does appear to be the only way, though.

Anyway, restricting the pointers doesn't appear to add any value to the optimizer, so don't have a huge appetite for it today. WDYT?

charlesnicholson: Oh, that's interesting; do newer C standards quality the printf parameters with restrict? If they do, then I'd be happy to make that change (with a restrict macro, for c++ compliance, as discussed).

You can also probe preprocessor definitions by passing them to a variadic macro that "counts" the number of values it has, e.g. https://gcc.godbolt.org/z/veGqT14Yj

stefano-zanotti-88: C23 does have the restrict qualification. See the synopsis in §7.23.6.1 of the standard.
Draft: https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf
Wikipedia has links to all the drafts for earlier versions, in the footnotes (of language revisions linked there):
https://en.wikipedia.org/wiki/C23_(C_standard_revision)

Here: https://en.cppreference.com/w/c/io/fprintf
it says that restrict was added in C99.
And here for C++ (which has no restrict anywhere, if I'm not mistaken):
https://en.cppreference.com/w/cpp/io/c/fprintf

Your macro magic is quite interesting. Just as evil as mine, I think.
It seems it could be gamed as well by some exotic values.
It's also quite crazy how it compiles cleanly at all, by just discarding arguments that would not be well formed before they can cause problems.

charlesnicholson: Yes, I agree that our macros are equally evil :) C has frustrating limitations for sure.

charlesnicholson:
Merged as https://github.com/charlesnicholson/nanoprintf/pull/321

PR #301, opened 2025-05-04
Fix: negative 0

charlesnicholson:
This PR is a reimplementation of https://github.com/charlesnicholson/nanoprintf/pull/290 in my own branch for easier building + testing. @stefano-zanotti-88 FYI!

### DIFF
--- a/b
+++ b/b
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-PYTHON3=$(command -v python3 || true)
+PYTHON3=$(command -v python3.13 || command -v python3 || true)
 if [[ -x "$PYTHON3" ]]; then
     "$PYTHON3" build.py "$@"
 else
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -190,10 +190,13 @@ NPF_VISIBILITY int npf_vpprintf(
 
 #if defined(__clang__) || defined(__GNUC__) || defined(__GNUG__)
   #define NPF_NOINLINE __attribute__((noinline))
+  #define NPF_FORCE_INLINE inline __attribute__((always_inline))
 #elif defined(_MSC_VER)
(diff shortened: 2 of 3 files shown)


stefano-zanotti-88:
Great!
From the size reports, it seems this shaved off a few bytes (if floats are enabled), on Cortex.
This is expected, as `< 0` has to do a bunch of extra stuff (`nan < 0` is false, and `-0 < 0` is false too), whereas now we only extract a single bit.
Would it also be possible to print stack usage information? The function-local usage, and the maximum usage through nested calls starting from each function.

charlesnicholson:
I've always been idly curious about stack usage, but haven't ever taken the time to measure it :) FYI I've often sacrificed stack usage for code footprint -- the format spec struct uses wider types than is necessary. I used bitfields originally but the cost of all the shifting and masking made a big differece vs a naked load.

But yeah, I'd love to profile it in the size_reports somehow.

stefano-zanotti-88:
I like a balanced approached about code-vs-stack -- not only is ram usually scarcer than flash memory, but for the code, you pay once; whereas for the stack, you pay once per thread. On embedded, this is annoying when temporarily debug-logging from threads that are otherwise lightweight, as I often forget to temporarily increase their stack size, and cause them to crash.
Anyway, npf is already low on stack usage compared to other libraries, and that's great.

`-fstack-usage` (at least with GCC) produces `.su` files with relevant information.
It seems clang has it too: https://reviews.llvm.org/D100509
I inspect them through the IDE, so I've never looked inside them nor parsed them into a nice tree-like structure myself.
Looking now at some example files, the format seems extremely simple, but the call graph is not there, it must be reconstructed from other compilation artifacts.
Maybe this could be useful:
https://mcuoneclipse.com/2015/08/21/gnu-static-stack-usage-analysis/

PR #304, opened 2025-05-04
Config flags for alternate form specifier `#`

charlesnicholson:
This PR is a reimplementation of https://github.com/charlesnicholson/nanoprintf/pull/288 that also updates the mpaland tests and the size reports.

cc @stefano-zanotti-88


charlesnicholson:
Takes 64 bytes off of the min-spec build!

PR #306, opened 2025-05-05
try windows mingw

charlesnicholson:


charlesnicholson:
@stefano-zanotti-88 - this is almost passing but i'm not sure what's going wrong when it tries to run the linked test executables. It's not giving a file-not-found error or any other output, but I suspect that it's a cmake / msys / bash pathing issue, since all of the intermediate .o files are being stored elsewhere in the build tree.

charlesnicholson:
I'm fine with fixing up the tests- it helps me get a deeper understanding of the changes, which helps me :)

I do my daily work from a Mac most days but have a Win11 beelink at my bench since the EEs use Windows. Next time I'm up there I'll play with the executables myself. There aren't any interesting details in the logs about attempting to run the executables; you'd think that they would report a seg fault or failed tests or something, so I suspect that CMake can't find the executables in the place it thinks they are, and that's causing an early failure with no output. But... who knows; it's all opaque until I investigate directly.

PR #308, opened 2025-05-05
Honor field-width / precision for %c and %s

charlesnicholson:
It's undefined by the spec but less surprising this way, and less code. This PR is a reimplementation of https://github.com/charlesnicholson/nanoprintf/pull/292 by @stefano-zanotti-88 with some test fixup.


charlesnicholson:
Size reports have some fluctuations- cortex-m0 seems to have gotten a few bytes larger but cortex-m4 seems a few bytes smaller. This is often due to inscrutable compiler micro-optimizations; I suspect that the extra prec = 0 lines caused it. I don't mind it, though. These kinds of things tend to grow and shrink a bit as new versions of gcc arrive.

Issue #132, opened 2022-01-19
Overhaul test system


charlesnicholson:
Rework the testing system to compile 100% of the tests against the 2^N flag matrix as well as C and C++ axes.

charlesnicholson:
From https://github.com/charlesnicholson/nanoprintf/pull/316#issuecomment-2869917935

===

I don't have an appetite for removing C++ from the tests entirely; a great deal of embedded systems code is written in C++ these days, and many users define NANOPRINTF_IMPLEMENTATION and include the header file in c++ source files. Nanoprintf must work in C++, and its tests must pass in C++! (I try to avoid C++ in embedded systems; it is the most complicated programming language ever designed by humans. Lots of folks like C++ firmware, and it provides conveniences over C in host-side code, so I support it..)

That said, you point out a very valid issue- nanoprintf has weak testing for pure C. I have had https://github.com/charlesnicholson/nanoprintf/issues/132 open for quite as while, as you have noticed :)

I would happily accept a PR that hits the following notes:

Get rid of doctest entirely.
Replace / rewrite the tests in a pure C test driver that can be compiled as C++ as well.
Can do the (2^N) * 2 (one for C, one for C++) exhaustive test compilation and execution, so that every test can be run in both C and C++.
I would love to change the CMake setup such that instead of today's approach of creating 2^N binaries (and invoking the linker that many times), it stamps out (2^N) * 2 files that all get compiled and linked into the same executable, and all run in one invocation. Doctest gives you that automatically via some static registration magic; it "finds" all of the tests via what I'm assuming is the standard C++ "define a class at file scope whose constructors register the tests with a central runner" technique. Perhaps the equivalent in this "C and C++" world is to also simply generate a file that explicitly calls x_y_z_tests or something for every x, y, z etc in the 2^N configuration. Perhaps there are other ways as well, but the heart of the approach, I think, needs to be along the lines of "generate a file that contains the 2^N set of NANOPRINTF_ define values currently being iterated over, that also includes a common .inl file that defines all of the tests" for the axes of [(C x C++), (All 2^N flag combinations)]. The executable would be large, containing a LOT of static copies of nanoprintf compiled with the _STATIC flag enabled, but I don't think that's a big deal even if the executable hits a ridiculous 64MB; it's just host-side testing.

This is a medium-sized project but I think it's worth doing; it's got a ton of wins:

It gives users and implementors 100% confidence that new code works across all languages and flags
[...]

PR #313, opened 2025-05-08
added test checks about the length of strings

stefano-zanotti-88:
Adds checks about the length of the produced string, and the size reported as the return value.
Also checks it against the return value of the sys printf.

This is just an idea. I haven't actually compiled the NPF tests (usual problems).
It should also be extended to other test files.

Adding these checks caught some issues. I've run those checks on (a replica of) conformance.cc only, not on the test cases in paland.cc nor the other test suites.
Issues:
`require_conform(output, "%c", i);`: prints "", returns 1
`require_conform("", "%+c", 0);`: prints "", returns 1
`require_conform(buf, "%p", p);`: due to implementation-defined behavior; might be fixed in your latest commits (about optional "0x" in %p), but we should check
`require_conform("0.00390625", "%.8Lf", (long double)0.00390625);`: my own sys printf prints "0.00000000". Maybe something to look out for when comparing npf to sys.
`require_conform("-0.00390625", "%.8Lf", (long double)-0.00390625);`: my own sys printf prints "0.00000000". Maybe something to look out for when comparing npf to sys

Note that printing '\0' is UB.
I've seen this behavior:
```
printf("a%cx", '\0'); // NPF: "ax", returns 3
printf("a%cx", '\0'); // GCC/clang: "ax", returns 3
printf("a%cx", '\0'); // other: "a", returns 3 (presumably actually writes "a\0x" in the buffer)
```

A test like `printf("a%cx", '\0')` should be added to our tests -- I'm thinking of a new file gathering all UB/IB, maybe the nan test cases could be moved there as well.

### DIFF
--- a/tests/conformance.cc
+++ b/tests/conformance.cc
@@ -38,19 +38,21 @@ namespace {
 void require_conform(const std::string& expected, char const *fmt, ...) {
   char buf[256];
 
+  int n_sys;
   std::string sys_printf_result; {
     va_list args;
     va_start(args, fmt);
-    vsnprintf(buf, sizeof(buf), fmt, args);
+    n_sys = vsnprintf(buf, sizeof(buf), fmt, args);
     va_end(args);
     buf[sizeof(buf)-1] = '\0';
     sys_printf_result = buf;
   }
 
+  int n_npf;
   std::string npf_result; {
     va_list args;
     va_start(args, fmt);
-    npf_vsnprintf(buf, sizeof(buf), fmt, args);
+    n_npf = npf_vsnprintf(buf, sizeof(buf), fmt, args);
     va_end(args);
     buf[sizeof(buf)-1] = '\0';
     npf_result = buf;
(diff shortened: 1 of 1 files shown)


charlesnicholson:
Thanks, this is a very good insight and I agree that it should be added to the test harness and a non-stupid solution should be applied (clearly npf is screwing up by returning 3). How would you like to proceed on this PR, given the conversation about testing in https://github.com/charlesnicholson/nanoprintf/issues/132 ?

(Additionally, I agree that your extremely thorough listing of all the UB issues should be extracted into an explicit UB test module and pulled out of the conformance stuff; how do you see this working with that?)

Maybe we should convert this PR to an issue instead to track it there?

stefano-zanotti-88:
There are tests for UB in #298. However, they are specifically to check that the "safety" option catches them correctly.
If that option is not used, NPF does something (crashing, or producing IB results), which is very much worth testing as well.
I'd like to have one single file for testing all standard-conforming behavior, and another file for IB behavior (plus all the other files like the various flavors of ftoa). Maybe paland should be removed, ie its tests split between the standard-conforming tests, and the IB tests.
And, the "vs sys printf" tests could be done in yet another file -- at most, the standard-conformance tests could be done (in the very same test case) both against an oracle and against the system printf.

About the problematic behavior of NPF:
%c with argument '\0' should just ignore that character. Emitting a '\0' mid-sequence and then emitting other stuff is just asking for trouble. It's true that the callback version could handle that '\0' without problems, but this would introduce an inconsistency between the callback and the sprintf versions that I'm sure is no good at all.
So, printf("a%cb", '\0') should emit "ab" and return 2

About %p, we should not test against sys-printf at all. And we should either:
* test with `(void*)some_integer_constant`, disregarding the fact that it might be technically UB to do this cast (but printf does nothing with it except cast it to an integer and print it, no actual pointer stuff involved)
* or, use real pointers, and test the pointer against a scanf of the produced string.
The second one might be "cleaner", but it's also more complex, and doesn't let us check for corner cases at all. Also, it makes tests completely irreproducible on desktop OSes, due to ASLR.

Feel free to organize things as you see most fit, ie open an issue about this, or else.

Issue #19, opened 2019-07-05
Float hexponent support: "%a" / "%A"


charlesnicholson:
Does anyone actually use this? I don't get it.

```
converts floating-point number to the hexadecimal exponent notation.

For the a conversion style [-]0xh.hhhp±d is used.
For the A conversion style [-]0Xh.hhhP±d is used.
The first hexadecimal digit is not 0 if the argument is a normalized floating point value. If the value is ​0​, the exponent is also ​0​. Precision specifies the minimum number of digits to appear after the decimal point character. The default precision is sufficient for exact representation of the value. In the alternative implementation decimal point character is written even if no digits follow it. For infinity and not-a-number conversion style see notes.
```

charlesnicholson:
Either way this should do exactly what %e does but just pass base 16 to `npf__ftoa_rev` or `npf__etoa_rev` or whatever. `npf__ftoa_rev` shouldn't take a `base` if `npf__etoa_rev` ends up being necessary.

stefano-zanotti-88:
A few thoughts about the why of %a.

The C23 standard says:
```
if the precision is missing and FLT_RADIX is a power of 2, then the precision is sufficient for an exact representation of the value
```

Also based on this, my impression is that %a can be used to serialize/deserialize a float value with perfect precision, in a bitsize-agnostic way.
When serializing/deserializing floats in decimal representation, we have rounding problems. Even if both sides properly ensure perfect rounding, it is still hard -- lots of code, and slow.
Instead, with %a it is quite straightforward to do.
Even faster and simpler methods of serializing/deserializing floats involve sending the binary representation itself, or a (hexadecimal) string representation of that. This, however, requires the deserializer to know the bitsize of the original float, whereas the output of %a doesn't require this.
Moreover:
```
converted in the style [-]0xh.hhhhp±d, where there is one hexadecimal digit (which is nonzero if the argument is a normalized floating-point number and is otherwise unspecified) before the decimal-point character […]
Binary implementations can choose the hexadecimal digit to the left of the decimal-point character so that subsequent digits align to nibble (4-bit) boundaries
```
leaves freedom to the implementation to use a representation that is not canonical nor the shortest, but is the most straightforward and faster to compute.
[...]

PR #315, opened 2025-05-11
Support for %a (and tangential fixes)

### LINKED ISSUE #295: Wrong formatting for '%0.*' with negative precision
The C23 standard says:
```
For d, i, o, u, x, and X conversions, if a precision is specified, the 0 flag is ignored. For other conversions, the behavior is undefined.
```
But also:
```
The precision takes the form of a period (.) followed either by an asterisk * (described later) or by an optional nonnegative decimal integer; if only the period is specified, the precision is taken as zero.
```
```
A negative precision argument is taken as if the precision were omitted.
```

This is the output produced by NPF:
```
printf("%010.*i", -1, 2); // "         2"
```
and this is by other implementations
```
printf("%010.*i", -1, 2); // "0000000002"
```
The bug only appears when a width is specified too, otherwise the '0' flag has no visible effect anyway.

NPF already ignores negative precision values, but in this specific case it ignores it but also ignores the '0' flag (as would be done if the precision was non-negative).

Note that GCC warns about calling printf with such arguments:
```
'0' flag ignored with precision and '%i' ms_printf format [-Wformat=]
```
But the warning is bogus: it does not understand that the precision is dynamic, and that it must be ignored if negative (it can't check it in the general case, but in this case it could see that the parameter is the compile-time constant -2).

The problem is that NPF discards the '0' flag immediately as it detects that some precision is specified. It does that before it has extracted the precision argument (in case of ".*"), so it cannot detect that the precision is actually "unspecified" (NPF completely ignores a negative precision, so it behaves --or should behave-- as if no precision was specified at all).
This happens on line 411:
if (out_spec->prec_opt != NPF_FMT_SPEC_OPT_NONE) { out_spec->leading_zero_pad = 0; }

I'm not sure how you want to go about fixing this. You need to skip this instruction, but you need to reintroduce it later (with modifications) in multiple places -- here you have one single place that acts for all integer specifiers, while later on you have one case for signed and one for unsigned integers.

Also, a similar problems happens with floats:
printf("%010.*f", -5, 1.234); // "001.234000" in other implementations, "0000000err" with NPF
The negative precision causes some overflow in the calculations, and although this is UB too, hence acceptable, it would be nicer to ignore a negative precision. However, a missing float precision must result in the default precision being used (6), which is done around line 415, but this too is skipped if the precision is present (and we don't know yet that we should ignore it).

One solution would be to always fetch the width/precision arguments immediately, instead of waiting for a successful parsing of the conversion specifier. This allows us to know already (on line 411/415) whether the provided precision had to be discarded.
Another solution could be to leave everything as is, since it is all UB; this might result in a smaller code footprint that any other "more consistent" solution.

stefano-zanotti-88:
This adds support for %a (see #19).
The support is (hopefully) fully compliant. The standard leaves a lot of leeway regarding %a. I've decided on a format that is 1) the one requiring the least code, I think; and 2) what everyone else seems to be doing as well (GCC/clang at the very least; including corner cases). So, we should have sys-printf-conformance, even though that wasn't the main goal.

I think the best way to enable %a is what I've already implemented: `NANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS` enables the parsing of %aefg. Then, individual options for %aefg must be enabled.
The default is to enable %f only.
When float is supported, but a specifier is disabled, that specifier will parse and fall back to a sensible alternative (among the other float specifiers) -- with a fallback of the type:
f->[g, e, a]
e->[g, f, a]
g->[e, f, a]
a->[e, g, f]
the first that is available.
Of course, this will be fully functional only when %e and %g support is added to NPF -- now %e and %g are forced to be aliases for %g.
When a specifier is disabled, its enum value is stripped out as well; the specifier will parse as if the fallback was parsed. That is, the aliasing is in the parsing of the specifier, not in the handling of the output generation.
This hopefully removes a little branching in some cases.
This "wrong" parsing does not create any problem, not even when adding safety checks (#298, to be updated), since %aefg are identical when it comes to parsing and checks.

I have kept the %a formatting (in `npf_atoa_rev`) completely separate from `npf_ftoa_rev`, since this is just a prototype for you to inspect. Most probably, all the float-related functions should share the same the nan/inf checks at the beginning (including the extraction of exp/mantissa, and the preprocessing for subnormals and exp unbiasing), and the "err"-emitting code at the end.
That is, some common code should handle those cases, and call into the proper ftoa/atoa/... function for the bulk of the work, passing the exp/mantissa as arguments. Such function should return an error (maybe len < 0) so that the common code knows when to produce "err".

Also note that I've changed the default precision for 'a' to the proper value.

The %a conversion in NPF uses the same types as used for %f (`npf_double_bin_t` `npf_ftoa_exp_t`, etc.).
Right now, it uses `npf_double_bin_t` for the mantissa calculations, so it always has full precision. Maybe dynamic scaling should introduced similarly to %f, though I feel that %a is more about accurate representation (and is faster already anyway), so I'm not sure whether dynamic scaling is appropriate. Maybe the user should specify two integer types, one for %e/f/g and one for %a (which defaults to full precision), though having two parameters might be too much configuration.

I haven't added dynamic integer scaling yet, because I need your opinion on that.
First, please read my notes about %a -- there's a new dedicated file; I'd keep that file as a container for various notes on the algorithms used, implementation choices and tradeoffs, standard conformance, etc; anything that's too in-depth to be in the readme. It would probably be better as a markdown file rather than the plain text I've used.
The matter is this: in my opinion, the only reason why %a exists is to have an easy way to get a round-trip string representation of floats, easier and faster and more accurate than the decimal representation.
So, using smaller integers and thus losing precision makes little sense. Then of course %a also supports padding, precision, etc., but the main usage is plain "%a" (with precision omitted), which ensures full precision.
I haven't used `npf_ftoa_man_t` to handle the mantissa, but plain `npf_double_bin_t`.
If you would still like to have dynamic scaling, part of the math should be changed to use `npf_ftoa_man_t` (or, as said, a different, user-selectable `npf_atoa_man_t`).
A lesser problem is the actual representation of bits in the reduced integer.
Now the handling is easy, as we basically have a number [000B_bbbb_bbbb_...], with B as the most significant bit (the implicit 1), and then other bits arranged in nibbles, each then converted to a char. In case of rounding, it can turn into [0010_0000_0000_...], without any overflow.
With reduced integers, we don't want to lose another 3 bits, so we would keep it as [B_bbbb_bbbb_...]; we would need to handle the rounding overflow properly, and also handle the high digit/char in a special way, and then proceed to the other nibbles, which can all be shifted into a known location (the highest one can't, without losing low bits).
So, nothing too complicated, but the code might grow, even though it would be offset by not having to handle large integers, on small systems.

A really tiny system might choose to enable %a and not %f, so that it will still be able to print floats (for machine consumption rather than human consumption), with much less code.
Although, a very constrained system could just print the hex representation of the underlying bit pattern of the float, with even less code.
Also, such a system might be interested in having support for %a but not for precision, which is not possible at the moment (float specifiers require precision to be enabled) -- but this might add complexity to the source code for little benefit.

I've added extensive tests as a separate file (not using doctest), that must be run by itself (it has its `main()`).
You might want to integrate them with the other tests.
I only tested it on a platform with `sizeof(double) == 8`, but I've written  the tests to also support `sizeof(double) == 4` (this hasn't been actually tested).
`sizeof(double) == 2` is not handled in the tests, though NPF does handle it -- NPF adapts to `DBL_MANT_DIG` etc.
I've used tests with the macro `PRC` (also checking against the system printf) and tests with the macro `PR` (only against an oracle string).
This is because my system printf has various bugs/issues (long double, nan/inf when precision is set, rounding mode, …).
However, most tests are implementation-dependent (as noted in the code), so you might want to change which cases actually test against the system printf.

Additional tests should be added to check the proper working of the fallback. For instance, something minimal like this:
```
#if !NANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS
  PR("%.0a"    , "%.0a", 0.0);
  PR("%.10a"    , "%.10a", -0x1.2345p10);
#else
  #if NANOPRINTF_USE_FLOAT_HEX_FORMAT_SPECIFIER
    PR("0x0p+0"    , "%.0a", 0.0);
    PR("-0x%1.2345p+10"    , "%.10a", -0x1.2345p10);
  #else
    PR("0"    , "%.0a", 0.0);
    PR("-1165.0781250000"    , "%.10a", -0x1.2345p10);
  #endif
#endif
```

***

The code has a couple TODO's about small things to be decided upon. Let me know about them.
Again, the current behavior is one of the easiest (so, less code), though not the absolute easiest; and it mostly agrees with GCC/clang.

***

Mind that this PR also fixes #295, ie the problem with negative precision passed as argument with `".*"`.
The fix was moving (and tweaking) the `prec`/`prec_opt` handling after the extraction of the `".*"` argument.

***

It also fixes an unreported problem (which is in a dark area of possibly-UB, but now it's "obvious" rather than "surprising"):
```
printf("%0*.3f", -10, 1.234); // before "1.23400000", now "1.234     "
```
The problem is: the '0' flag must be ignored if '-' is used.
Before, '0' was explicitly cleared when '-' was encountered. But this missed the case of '-' set as a consequence of a negative width.
Now, I've removed those explicit clear instructions, and added a condition at the end (when actually printing the padding).
The problem was even more severe with %a:
```
printf("%0*.3a", -12, 0x1.234p1); // before "0x1.234p+100", now "0x1.234p+1  "
```

The fix is a single line change:
```
if (fs.leading_zero_pad) {
```
to
```
if (fs.leading_zero_pad && !fs.left_justified) {
```
in the code that computes `pad_c`; plus a little cleanup of no-longer-needed clearing of `leading_zero_pad`.


charlesnicholson:
This looks great, and I'd love to merge it. I have a few observations-

1. I think I made a mistake in treating %e, %g, and %a as float; I'd rather that they just behave identically to other configuration-disabled flags and just print the format string instead. That would let us delete a bunch of boilerplate macro stuff in this code- if "float" is enabled, you get %f and %a, and %e and %g end up just printing those format strings to the output.
2. The tests are excellent and exhaustive, but figuring out how to incorporate them into the build in light of the conversation at https://github.com/charlesnicholson/nanoprintf/issues/132 - I don't mind treating this the way we've done the other ones, where I copy the PR code into my PR and rewrite your tests in terms of doctest. Please let me know!
3. Examining the impact on binary size, since this is a bunch of new code. Maybe there are optimizations lurking.

Let me know how you'd like to proceed. If you want me to take nanoprintf.h and the new test file and integrate it myself, could I ask you to resolve conflicts and bring your PR up-to-date with main please?

stefano-zanotti-88:
1. Your choice of %ega being aliases for %f is debatable. Either way it's non-standard, but printing a number as if with %f lets the user see the number somehow, rather than seeing the unhelpful "%a" or whatever.
Sure, they should just change it to use one of the available specifiers, but maybe they can't (they want to switch between npf and a standard printf at will, or they would have to refactor lots of code).
Also, using a %f representation in place of %e causes no problems; that would not be for treating "%llu" as "%u", etc.
So, I found your initial choice very reasonable, though the other one of not even parsing it is sensible as well.

Anyway, let me know how you'd prefer to proceed about this. Having the aliases results in larger code; but that's only for having multiple cases in the switch that writes to `conv_spec` (multiple cases that jump to the same body), so the difference can't be more than a bunch of bytes.

2. Indeed, I'd rather let you redo this and any tweaks in your own PR.
As said, I'm comfortable and fast testing with this minimal setup (as in the tests_a.c file), but of course it should be integrated with the other tests, and right now I can't fully handle that part since I can't compile with all the CMake machinery.

3. Please look at the TODOs in the code for a couple small things. Also, there is common code in ftoa and atoa that most certainly needs to be factored out.
One thing I've just seen (will fix it later) in atoa:
```
if (spec->prec < 0) {
```
is now redundant, as `prec` is already properly set. This comes from earlier attempts on my part.
Another thing: converting a nibble to a hex digit might be factored out -- it is used by %a and by %x. It's certainly much slower to move it to a function, but maybe it's a bit smaller? It's quite dubious.

4. I'm happy to bring it up to date, but there's a little "management problem" due to my little familiarity with the whole PR stuff. #287 was done from my main branch. I haven't touched that nor updated that to your own latest main, since I don't know how that will affect #287. Ideally, I would just erase my main branch, recreate it to track yours, and then only work on feature branches for the PRs. But, can I move the PR to another branch (newly created, to point to the same commits)? Or, should I just park my main branch there, and track your main branch with another new one?
Also: if I were to delete branches or even the whole repo, would that impact npf's PRs / discussions, or would the relevenat diffs be preserved, so that looking at the code in PRs would still make sense?

charlesnicholson:
(BTW i'll take a shot at pulling in this PR ASAP, maybe over the weekend, but sometimes i can do them at night)

stefano-zanotti-88:
I've brought this branch up to date with your main branch.

I also added a missing error-check about NANOPRINTF_USE_ALT_FORM_FLAG being undefined.

This branch still contains fixes for "0-" coexistence and for prec/prec_opt handling (to properly ignore negative precisions given with `.*`)

And there are still TODOs in the code about minor choices to be made.

Let me know if you want to work from here, or prefer to have those fixes extracted into their own PRs.

PR #316, opened 2025-05-11
C-compatible tests

stefano-zanotti-88:
Related to #132.
Can you help me understand what advantages there are in using doctest rather than just defining custom checks?
Though doctest claims to be very fast, it's certainly still heavier that a more trivial framework for NPF tests, especially since we need to compile them multiple times, for the various combinations of the options. I think CI could give results quite faster if we were to replace doctest.
Also, it requires C++, which is another source of heaviness, and prevents us from testing NPF in C -- unless we were to compile NPF in a .c translation unit (without static visibility), so that the tests can call into that even from C++; but this is more complicated, and I don't think it's the way to go.
Moreover, a custom test framework would make failure messages more expressive, for instance by printing the line number that is failing, and other contextual information.
Maybe that can be done in doctest too, I don't know, but it seems it's not being done right now.
When debugging a single test case, the more verbose messages are not that useful. But when looking at multiple failing cases, it might be more insightful to see all of the failures together, than to have to step through each of them.
Most importantly, though, CI errors would have important information about the cause of the error, without having to reproduce them locally (and possibly with a different setup, so with hard to reproduce issues).

This PR is an example of how we could implement such a test framework, which can be very minimal in fact.
Basically, I replaced the REQUIRE macro with some custom checking/printing/aborting; and I wrapped all the test cases in a `main()` function. Of course only one file can be run at a time, whereas now you can maybe run multiple files using doctest.
Then I trivially replaced C++-specific stuff like lambdas and `nullptr`.

Running the conformance.cc file found a few bugs (already reported elsewhere), that were not detected by npf-vs-sys string comparison:
* %c with '\0' argument
* %p can't be checked against system printf, since it's IB

With C:
```
ret!=len: 1 != 0; for "" (exp "")
  -> Fail @160 (originated @194): ""
ret!=len: 1 != 0; for "" (exp "")
  -> Fail @160 (originated @198): ""
npf!=sys: "0xeb339dfb58" "000000EB339DFB58"
  -> Fail @156 (originated @461): ""
npf!=exp: "0xeb339dfb58" "000000EB339DFB58"
  -> Fail @157 (originated @461): ""
ret!=sys: 12 != 16; for "0xeb339dfb58" vs "000000EB339DFB58" (exp "000000EB339DFB58")
  -> Fail @161 (originated @461): ""
npf!=sys: "                  0xeb339dfb58" "              000000EB339DFB58"
  -> Fail @156 (originated @467): ""
npf!=exp: "                  0xeb339dfb58" "              000000EB339DFB58"
  -> Fail @157 (originated @467): ""
npf!=sys: "0.00390625" "0.00000000"
  -> Fail @156 (originated @614): ""
npf!=sys: "-0.00390625" "0.00000000"
  -> Fail @156 (originated @616): ""
ret!=sys: 11 != 10; for "-0.00390625" vs "0.00000000" (exp "-0.00390625")
  -> Fail @161 (originated @616): ""

Tests failed: 10 / 2408. FAIL.
Done.
```
With C++:
```
ret!=len: 1 != 0; for "" (exp "")
  -> Fail @160 (originated @194): ""
ret!=len: 1 != 0; for "" (exp "")
  -> Fail @160 (originated @198): ""
npf!=sys: "0" ""
  -> Fail @156 (originated @375): ""
ret!=sys: 1 != 0; for "0" vs "" (exp "0")
  -> Fail @161 (originated @375): ""
npf!=sys: "0xd31103f798" "000000d31103f798"
  -> Fail @156 (originated @461): ""
npf!=exp: "0xd31103f798" "000000d31103f798"
  -> Fail @157 (originated @461): ""
ret!=sys: 12 != 16; for "0xd31103f798" vs "000000d31103f798" (exp "000000d31103f798")
  -> Fail @161 (originated @461): ""
npf!=sys: "                  0xd31103f798" "                    d31103f798"
  -> Fail @156 (originated @467): ""
npf!=exp: "                  0xd31103f798" "                    d31103f798"
  -> Fail @157 (originated @467): ""

Tests failed: 9 / 2409. FAIL.
Done.
```
I've marked the failing cases with `// TOFIX`

I wanted to uncover possible C-vs-C++ bugs in NPF. Instead, it uncovered bugs/inconsistencies in my local printf:
line 375: only in C++, `printf("%#.0o", 0)` results in `""` instead of `"0"`
line 467: %p is different under C/C++
line 614-616: only in C, it doesn't support long double (though it's C11)

***

To compile the tests as C/C++: I haven't tried it (I just changed the file extension, to see that noth C and C++ work), but in GCC you can use `-x c` or `-x c++` to force a specific language, regardless of the extension.
The exact same option should available in clang.
In MSVC, there are `/Tc` for C, and `/Tp` for C++.

### DIFF
--- a/tests/conformance.cc
+++ b/tests/conformance.cc
@@ -13,9 +13,13 @@
 #define NANOPRINTF_IMPLEMENTATION
 #include "../nanoprintf.h"
 
-#include <string>
+#if defined(__cplusplus)
+  #include <string>
+  #include <cmath>
+#else
+  #include <math.h>
+#endif
 #include <limits.h>
-#include <cmath>
 
 #if NANOPRINTF_HAVE_GCC_WARNING_PRAGMAS
   #pragma GCC diagnostic push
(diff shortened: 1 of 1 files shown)


charlesnicholson:
Hi @stefano-zanotti-88 -

I chose doctest because it's what I use at work and in other open-source projects. Its greatest value IMO comes from its almost-arbitrary expression-and-capture parser that exists only in C++ (even when testing C functions), and its `SUBCASE` isolation magic where each `TEST_CASE` is re-run once per `SUBCASE` and skips the other subcases present.

That said, nanoprintf doesn't really use either of those features too strongly, other than being able to take advantage of the `==` operator for `std::string` by performing checks like `REQUIRE(std::string{"blah"} == function_that_returns_char_array_holding_blah);`

I don't have an appetite for removing C++ from the tests entirely; a great deal of embedded systems code is written in C++ these days, and many users define `NANOPRINTF_IMPLEMENTATION` and include the header file in c++ source files. Nanoprintf must work in C++, and its tests must pass in C++! (I try to avoid C++ in embedded systems; it is the most complicated programming language ever designed by humans. Lots of folks like C++ firmware, and it provides conveniences over C in host-side code, so I support it..)

That said, you point out a very valid issue- nanoprintf has weak testing for pure C. I have had https://github.com/charlesnicholson/nanoprintf/issues/132 open for quite as while, as you have noticed :)

I would happily accept a PR that hits the following notes:
* Get rid of doctest entirely.
* Replace / rewrite the tests in a pure C test driver that can be compiled as C++ as well.
* Can do the (2^N) * 2 (one for C, one for C++) exhaustive test compilation and execution, so that every test can be run in both C and C++.

I would love to change the CMake setup such that instead of today's approach of creating 2^N binaries (and invoking the linker that many times), it stamps out (2^N) * 2 files that all get compiled and linked into the same executable, and all run in one invocation. Doctest gives you that automatically via some static registration magic; it "finds" all of the tests via what I'm assuming is the standard C++ "define a class at file scope whose constructors register the tests with a central runner" technique. Perhaps the equivalent in this "C and C++" world is to also simply generate a file that explicitly calls `x_y_z_tests` or something for every `x, y, z` etc in the 2^N configuration. Perhaps there are other ways as well, but the heart of the approach, I think, needs to be along the lines of "generate a file that contains the 2^N set of `NANOPRINTF_` define values currently being iterated over, that also includes a common .inl file that defines all of the tests" for the axes of `[(C x C++), (All 2^N flag combinations)]`. The executable would be large, containing a LOT of static copies of nanoprintf compiled with the `_STATIC` flag enabled, but I don't think that's a big deal even if the executable hits a ridiculous 64MB; it's just host-side testing.

This is a medium-sized project but I think it's worth doing; it's got a ton of wins:
1. It gives users and implementors 100% confidence that new code works across all languages and flags
2. It speeds up the build - only linking once is much better than linking 2^N * 2 times
3. It speeds up the tests - running one optimized mega-test file will be faster than launching 2^N * 2 test processes, and leaves the door open for multithreading the tests if it ever matters.

If you have an appetite to do something like this, I think it would be great. It does mean figuring out your build environment and leaning into the build infrastructure, though; this one is big enough that I think anyone working on it (myself included) needs to have the build fully operational on their computer as they iterate.

(I'll capture this in an issue ticket as well, so it's not tied to the fate of this particular PR)

stefano-zanotti-88:
> its SUBCASE isolation magic where each TEST_CASE is re-run once per SUBCASE and skips the other subcases present.

I'm not sure I understand this. You mean that inside a test case you can define setup/teardown code, that gets automatically re-run for each subcase?

About C testing: my intention is not to test NPF only in C. It is to make the tests C-only. In this way, and if the test includes its own copy of NPF (I'm not sure if that's the case now, or all the tests use a global, external instance of NPF), then it can be compiled once as C, and once as C++, via compiler switches. This way you can test for C/C++ differences, knowing that the code for everything is identical in the two cases, and any difference is due to the language.

About your "notes":
* Get rid of doctest entirely: it's already in this PR: just define NPF_USE_DOCTEST as 0. Of course it can then be removed completely, I left it so that you could compare the two.
* Replace / rewrite the tests in a pure C test driver that can be compiled as C++ as well: this is already in place as well
Of course I've done it just for one file, to get your feedback, and would have to be done in other files as well.
Some common code about the test checks could be extracted to an .h file then.

About the (2^N)*2 tests: the following is basically what you're proposing too, isn't it?
* a script generates 2^N "config" files that define the various npf options with all possible combinations.
* each of these files includes one test file, which in turn includes NPF (with IMPLEMENTATION and STATIC)
* each of the test files has a main_<filename> function (with external linkage) to run the test
* we have a main file that calls all the test functions
* this can be repeated for C and C++, by having different names for the test function (eg "main_<filename>_c" and "main_<filename>_cpp") based on `#if defined(__cplusplus)`; and by having 2^N config files with .c extension, and another 2^N identical files with .cpp extension
In this way, one single main file will then call into all the tests. Every single combination will have its own npf static copy, and its own public test function, and the main will call all the functions, so that the whole test is just one run of one executable.

***

I don't think I have time to fully work on it myself, especially because of the build system issues that might arise.
I can probably try and get a prototype of the "generate all the files, set up the includes, build everything" machinery, and see if it works with my toolchain; but I can't test across toolchains, and it's not unlikely that there will be issues I can't solve.
Let's put it on ice as far as I'm concerned. Maybe I'll do try and piece together a prototype some time in the future.

charlesnicholson:
See https://github.com/doctest/doctest/blob/master/doc/markdown/tutorial.md#test-cases-and-subcases for a detailed explanation of how doctest supports traditional fixtures as well as the SUBCASE behavior I mentioned. It's quite elegant and powerful!

> About C testing: my intention is not to test NPF only in C. It is to make the tests C-only. In this way, and if the test includes its own copy of NPF (I'm not sure if that's the case now, or all the tests use a global, external instance of NPF), then it can be compiled once as C, and once as C++, via compiler switches. This way you can test for C/C++ differences, knowing that the code for everything is identical in the two cases, and any difference is due to the language.

To be pedantic, this means not making the tests C-only, since C and C++ are different languages and C++ is not a strict superset of C. This means writing the tests in a constrained version of C++ that is compatible with whatever version of C that we want the tests to compile against. This is what I meant.

I agree with the rest. I might take a shot at it- the build system already uses python3, which is a good language to do the file-stamping in, and I'm handy enough with CMake at this point to do it.

stefano-zanotti-88:
Yes, that's what I meant with my loose term "C-only".

To be even more pedantic, "A constrained version of C++ that is compatible with C" is also the same as "A constrained version of C that is compatible with C++", that is "A version that only uses features that are supported and identical (or almost so) in C and C++".

That said, I don't expect any problem from the test code itself, since we're not gonna use esoteric stuff.
And I don't expect any from NPF either, but that's better tested.

PR #324, opened 2025-05-15
Fix nul char

stefano-zanotti-88:
As discussed previously, this treats `'\0'` as if it was an empty string.
It is UB anyway, but at least now it generates consistent results -- though different from various sys printfs, which insert the char in the output, and report a length inconsistent with the strlen of the resulting string.


charlesnicholson (approved):
This looks great, thanks for the PR! Are you ready for it to be merged?

stefano-zanotti-88:
It's ready.

But you'll have to work on the tests for %c. I've left the non-UB ones, and added others as comments; they should be handled somehow, but the whole testing matter is still a work in progress, as of our other discussions

Issue #18, opened 2019-07-05
Float dynamic support: "%g" / "%G"


charlesnicholson:
Analyze the value and figure out whether to print float-decimal or float-exponent

Issue #17, opened 2019-07-05
Float exponent support: `%e` / `%E`


charlesnicholson:

charlesnicholson:
Maybe something cheapo like this would work:

Determine a multiplier of 10.0f or 0.1f based on initial condition of `0 <= abs(x) <= 1` or `abs(x) > 1`. Determine log10(x) by counting how many applications of the multiplier brings x into `1 <= abs(x) < 10`. Once x is exponent-normalized just print it like a normal reversed float, but stick the `e+02` (or whatever) reversed at the beginning of cbuf.

This means that the concept of precision will have to be pushed into this new `etoa_rev` function maybe?

charlesnicholson:
Also make sure that there aren't any floating point numbers where successive multiplications of 0.1 or 10.0 won't eventually normalize it to `1 <= abs(x) < 10`. Floats are hard.

PR #325, opened 2025-05-19
Support for %e and %g

stefano-zanotti-88:
This adds support for %e and %g.
See #17 and #18

This uses the same exact algorithm as for %f. It differs in the bookkeeping regarding how many digits are needed, where to put the dot, etc.

Basically, %f (already implemented in NPF):
- ignores the sign and padding (it's done outside)
- initially checks for nan/inf and handles them specially
- composes the output string in reverse, in the internal buffer
- extracts the "integer" and "fractional" parts of the mantissa. They can be more easily converted to decimal by iterating on increasing/decreasing powers of 10 (starting from the unit, which is `2**0 == 10**0 == 1`).
- converts the integer part to string as is done for any unsigned integer
- converts to fractional part to string by progressing through successive negative powers of 10
- already knows the precision, so it knows where the decimal point is (in the buffer)
- the integer part can be converted in the normal way (from the units up), progressing rightward in the buffer
- the fractional part can be converted "backward" (from tenths down), progressing leftward in the buffer
- the number (as whole) has a final rounding pass: the binary maths on the mantissa produces a carry (from the integral part, then overwritten by the fractional part, if it exists) which is used to round the last digit, and cascade the rounding upward (rightward). This may produce an extra digit.
- the fractional part is bound by the precision. The integral part may overrun the buffer, so there are checks in place. In case of overrun, the "err" string is produced.

Whereas, %e:
- ignores the sign and padding (it's done outside)
- initially checks for nan/inf, just like %f
- composes the output string in reverse, in the internal buffer
- it first composes the mantissa, similarly to what is done for %f (but see below), and only at the end it composes the exponent. This is because of rounding, which happens at the end of the mantissa conversion, and which might change the exponent too (its length, most importantly)
- the mantissa is composed as the integral part first, then the fractional part. No dot is added right now.
- the fractional part can end early (because it's limited by the precision)
- the integral part cannot end early (because we compose it from the units up, and don't know how many digits it has beforehand); this also means that the rounding can be based, in some cases, not on the normal carry used in %f too, but on both the carry and the already-emitted, to-be-discarded digits.
- the integral part can have trailing zeros. These are not emitted immediately (they may not be needed at all, and may "temporarily" overrun the buffer). They are emitted later, if needed (when we start looking at the fractional part)
- the fractional part can have leading zeros. We emit them only if needed, ie if we already have at least an integer digit
- the trailing zeros in the fractional part are only emitted at the very end (as a sort of padding)
- the integral mantissa is composed rightward, starting from the beginning of the buffer. As soon as it is done, it is moved to the end (right) of the buffer, so that the fractional mantissa has all the necessary room
- when the int+frac mantissa is done (plus final rounding), we also know the exact exponent
- we check whether the string fits in the buffer -- we know the exponent, the precision, and the whole mantissa, so we can indeed check that
- we compose the exponent as a normal unsigned integer + sign (at least 2 digits, even if it's within [-9, 9])
- we add 'e'
- we add the missing trailing zeros in the fractional part
- we move the whole mantissa leftward (it is still at the end of the buffer), except the most significant digit
- we add '.' if needed
- we move the most significant digit

%g:
- everything proceeds as for %e, except the following:
- when we have prepared all the digits in the buffer, and rounded them, we also know the exponent that %e would have; so we can choose between %e and %f (see the standard rules, and the comments in the PR).
- we remove all the trailing '0's (implicit, or already put in the buffer) if needed, as well as the dot
- we format the digits appropriately, adding the necessary leading/trailing '0's (in the integral and/or fractional part), the dot, and the exponent
- as for %e, we have prepared the digit at the end of the buffer, and the final formatting happens at the beginning of the buffer, so that the string is ready to be output; digits only ever move towards the front of the buffer (or stay in the same place), so we never have any overwriting (corruption) in the buffer.

I've left 3 functions in the code, for you to experiment:
`npf_ftoa_rev`: the original NPF function for %f only
`npf_etoa_rev`: %e only
`npf_gtoa_rev`: %g (which then selects between %e-like and %f-like), but also %f and %e directly

For full compatibility, only `gtoa` is needed. But it has some complicated logic that I don't think the compiler can optimize as much as `ftoa` or `etoa` even though we do disable the unneeded code-paths when a specifier is disabled.
In particular, the code path taken to print as %f is not as optimized as the path in `ftoa`, since the general case can make less assumptions, and is structured differently enough that the compiler is probably unable to turn them into identical code, even when conditionally compiling out the non-%f parts.
My feeling is that: if only %f is needed, npf_ftoa_rev is better; if only %e is needed, npf_etoa_rev is best; in all other cases (g, ef, eg, fg, efg), npf_gtoa_rev is better. Here, "better" means the least amount of total code added. Regarding performance, npf_ftoa_rev and npf_etoa_rev are always better than npf_gtoa_rev (maybe on par, but I very much doubt the compiler can optimize it that much). In particular, %g must also think about removing trailing '0's, which is not a concern for %e and %f, and this behavior is active even if '#' is disabled ('#' allows to turn it off, otherwise it is on -- so that code can't ever be removed). And, %f has some optimizations because it already knows how many digits it will need and where they will end up (the fractional length does not depend on the actual value), while %e and %g cannot determine that until the digits have been produced, and this also means some moving around of digits inside the buffer.

***

This PR has working code for %e and %g, but:
- common parts of %e-%f-%g can be factored out. Specifically the handling of nan/inf, the handling of "err" (producing the string; whereas detecting the error is conversion-specific). I didn't do it because my intention is not to keep all of these functions, but just to let you play around with them in order to decide how to proceed
- `etoa_rev` and `gtoa_rev` probably have some superfluous variables, that could be tracked in other ways, or implicitly; the compiler might or might not be already optimizing them optimally
- `etoa_rev` and `gtoa_rev` might have impossible code paths -- which might be proved to never execute because of previous checks, or properties of the float layout and performed math; I'm thinking for instance about the interplay of carry, significant digits, and the precision (which makes us stop before enumerating all the possible digits)
- `etoa_rev` and `gtoa_rev` have some error cases in which they could bail out earlier, skipping work that would be wasted anyway. The additional checks for early bailout may or may not be worth it, as they might increase the code footprint; I haven't added all the possible checks in this regard, as my feeling is that it's better to have a slow function when "err" is produced, than to have larger code (and slightly slower functions when no error is produced)
- especially `gtoa_rev` leans more to the readability side, thus introducing various local variables, and not reusing older variables or recalculating values that can be deduced from others, which would make the code quite more obscure. Maybe the compiler already optimizes as much as possible, maybe not. For now, I think readable code is better for you to get to know what it's doing, and then it can be explored how to reduce/golf it.
- it needs a *thorough* review to check for correctness, especially regarding rounding.

Let me know your thoughts on all of this, and how you want to proceed.

***

Working on this, I found unexpected (for me) issues with precision. I'll mention it here, though you might want to discuss it in an issue of its own.
There is one failing test:
```
printf("%.15e", 0x1.ffffffffffffep+1023);
```
This should print: "1.797693134862316e+308"
But prints: "1.797693134862315e+308"
The problem seems to be intrinsic to our algorithm, which accumulates too much error over the many iterations required by such a large number (2**1023 handled by u64 -> 1023 - 52 = 971 loop iterations for the integer part).
Please see my notes in the info.txt file, especially about the error. I thought we could not accumulate so much error, but evidently we can.
The exponent scaling with u64 terminates with this value:
0b1111100101111010111000111101000011010010010001000101111000001100
which is then converted to digits as:
17976931348623154700
and rounded as:
1797693134862315
The digit/rounding part is correct. The problem is in the scaling.

I tried using u128 for the internal scaling variable, and the scaling terminates with a different value (of course), which at some point (a point where we can compare it with the u64 above) it gets:
0b1111100101111010111000111101000011010010010001000101111110001101
which is then converted to digits as:
17976931348623155085(6124328384506240271)
and rounded as:
1797693134862316

For comparison (u64 vs u128)
```
0b1111100101111010111000111101000011010010010001000101111_000001100
0b1111100101111010111000111101000011010010010001000101111_110001101
17976931348623154700
17976931348623155085(6124328384506240271)
```
We have ~9 bits of error.

I know such a small error can be acceptable for NPF, but I would like you to double check this and give your opinion. I really thought that we could have full precision if we used at least as many bits as in the original float.
I thought we had (N*log10(2) - 0.5) digits of precision, thanks to rounding.
This would be (64*log10(2) - 0.5) ~= 18.77, but we actually have less than 16 (1 integral + 15 fractional).

Of course we do round the number in successive iterations, and rounding more than once is bound to cause problems. However, it seemed that the rounding was "sparse" enough (separated by a division by 5), that it would not accumulate this much.

Experiments showed than, in this specific case, we would need 68 bits to get the correct result.
I'm thinking whether we should turn the carry into an 8-bit variable, so that we get 7 additional bits. Still not enough to correct the 9-bit error of this case, but enough to get 16 output digits right (I'm not sure if this would be enough for round-trip; the 17th digit might be wrong, and still ensure round-trip).
The carry arithmetic would get fairly more complicated.

If you use __int128_t for scaling too, you'll see that a bunch of other test cases fail: for now, I've tailored the oracle string to the value that NPF outputs with u64, which is round-trip, but not the most accurate; with __uin128_t, we do get the most accurate output.

Also note that this problem surfaced with %e, because we can print larger numbers with it. However, the same problem exists for %f, if we set a buffer size large enough for it not to print "err":
```
printf("%f", 0x1.ffffffffffffep+1023); // 17976931348623154700000...
```

What do you think of all of this?

Also please look at the new "notes.txt" file. It has some claims about the algorithm and its precision -- or what I can understand of it -- but they clearly need a review, especially because of these issues with precision.
Anyway, I think it provides a nice documentation of what NPF is doing, especially for future maintainers.


charlesnicholson:
@stefano-zanotti-88 thanks for the approach, I developed it further and landed it in #369

PR #327, opened 2025-05-19
Fixed handling of "0-"

stefano-zanotti-88:
This fix is extracted from #315, as it is a bug that needed fixing for that PR, but appears in other contexts too.
```
printf("%0*.3f", -10, 1.234); // before "1.23400000", now "1.234     "
```


stefano-zanotti-88:
I had to merge it with #328 for a proper fix. See commit https://github.com/charlesnicholson/nanoprintf/pull/327/commits/b457f9b566e25e8eeb2d0618aad488706e24189d

Then, I added a nicer default for %p, which is related to these padding issues. This is IB, so maybe you don't want to accept that, to save a few bytes of code.

stefano-zanotti-88:
Maybe I'll try with WSL2.
As for the fix, it seems there are still some issues with the various NPF option combinations, which I'm still trying to sort out.

stefano-zanotti-88:
I've fixed all the issues, and reverted the changes about %p -- they need a different PR
There are still failing cases. However, at least one of those cases should be revisited:
https://github.com/charlesnicholson/nanoprintf/blob/49628bd79977478dbd7c9a1b5277c44e5d49b118/tests/unit_parse_format_spec.cc#L151-L153
In my view, it is not correct to expect the parsing to produce `!leading_zero_pad`. The flag is there, and it can only be properly ignored after fetching the dynamic precision (if any), which is not part of parsing.
I think these semantic constraints should be left to the caller (as is done to the main caller of the parser, ie `npf_vpprintf`, and not addressed by the parser.

I haven't changed that test yet, I'll be waiting for your feedback.

charlesnicholson:
Ah, I see, I just left feedback on the %p PR you opened https://github.com/charlesnicholson/nanoprintf/pull/329 - I see the issue now, that the spec can't be resolved until the dynamic star args are read, and the latter only happens in the actual formatting code. I don't mind this change because of the nature of it, but it does mean changing or deleting existing unit tests that harnessed the previous logic of "npf_parse_format_spec only returns legal combinations".

stefano-zanotti-88:
Addressed by #340

PR #328, opened 2025-05-19
Fix handling of negative dynamic precision

### LINKED ISSUE #295: Wrong formatting for '%0.*' with negative precision

stefano-zanotti-88:
This fixes #295.
It was extracted from #315.
```
printf("%010.*i", -1, 2); // before: "         2", now: "0000000002"
```


stefano-zanotti-88:
#328: Yes, I realized that other PRs needed this as a prerequisite, and made the same changes there too.
This is all already covered by your #340, so I'm closing this.

PR #330, opened 2025-05-20
Removed spurious NUL terminator

stefano-zanotti-88:
NPF always writes a NUL character at the end of the buffer, to handle the case when the normal terminator was skipped because the buffer is full.
With this PR, it only writes a single NUL character, in the appropriate position.


nanoprintf.h
```diff
@@ -315,6 +315,7 @@ typedef struct npf_bufputc_ctx {
   #include <intrin.h>
 #endif
 
+#define NPF_MIN(a, b)    ((a) <= (b) ? (a) : (b))
```
charlesnicholson, line 318: I'd prefer a `static int npf_min` version of this, so that `a` and `b` aren't at risk of getting re-evaluated when called multiple times (that's what motivated the current `npf_max`). All compilers I've seen inline it; and I'd be happy to see the `NPF_FORCEINLINE` qualifier added to both. Would you mind making that change please?

charlesnicholson (approved):
Thanks for the PR! I was surprised to see this take 4 bytes off of the low-end CM0 and CM4 builds!

nanoprintf.h
stefano-zanotti-88: No problem.
However, the reason for the macro is that it works for all types (provided that you use it correctly). In this case, I need MIN for `unsigned int`.

What do you prefer? The macro as-is, a dedicated function like "npf_min_uint", or other?

charlesnicholson: oof, sorry, i missed that. I guess they're both fine as macros; we can go the other way, as long as the code never calls anything that is vulnerable to multiple evaluations.

charlesnicholson: I'd just prefer to have min and max look uniform, instead of having one as a function and the other as a macro :)

stefano-zanotti-88: I've done it.
As you can see in the diffs, NPF_MAX is only used in two places, with trivial expressions.

PR #331, opened 2025-05-20
Potential stack usage reduction

stefano-zanotti-88:
This is completely useless on most platforms, but will probably save some bytes and some code on 8-bit architectures (eg AVR), where int is >= 16 bit, so it is larger than the machine word.
Do we have a way to verify that we do have some gains?


stefano-zanotti-88:
I manually check the stack usage with GCC with `-fstack-usage`, and then inspect the `.su` file generated for a `.c` file like this:
```
#define NANOPRINTF_ENABLE_SAFETY_CHECKS
#define NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS    1
#define NANOPRINTF_USE_PRECISION_FORMAT_SPECIFIERS      1
#define NANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS          1
#define NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS          1
#define NANOPRINTF_USE_BINARY_FORMAT_SPECIFIERS         1
#define NANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS      1
...
#define NANOPRINTF_IMPLEMENTATION
#include "nanoprintf.h"
```

It does not use `NANOPRINTF_VISIBILITY_STATIC`, so that the functions are not optimized away. Maybe the same can be obtained with the current `static_nanoprintf.c`, we'd have to check whether some functions disappear. Probably just `npf_pprintf` and `npf_snprintf`, which are quite thin anyway.
I don't know how you could integrate that in the current CI.

I see no changes, but this is expected, on 64-bit platforms.

My question, however, was more about: is there a way to also build NPF with avr-gcc, as an example 8-bit architecture? This would provide size reports for that, as well as tests for cases that are otherwise not covered (see for instance #246)

charlesnicholson:
Yeah, sure, I'd love to harness nanoprintf on github with an 8-bit size report! I think it's straightforward enough, it's just a matter of installing it into the docker image that I use for the size-reports. It's here:

https://github.com/charlesnicholson/docker-images/blob/main/Dockerfile

Is installing avr-gcc as simple as just downloading it and un-archiving it, then running it? If you want to file an issue for it i'll do it when I have a minute...

stefano-zanotti-88:
I'm sorry, I don't know how to install avr-gcc. I've always used it through Atmel Studio (an IDE based on Visual Studio).
Maybe you also need avr-binutils (and possibly avr-libc), if they're not already bundled with the compiler

As an unrelated note, see here for the size of types:
https://gcc.gnu.org/wiki/avr-gcc#Type_Layout
`int` is 16-bit
`double` and `long double` are both 32-bit (unless you override this with a compiler option)

charlesnicholson:
No problem; it looks like for non-windows OS's you can just download and unzip it and it works like a normal gcc.

charlesnicholson:
BTW if you want to hit "Update branch" to merge the latest main into your branch, I'll approve + merge this PR! It's just not up-to-date yet.

charlesnicholson (approved):

PR #334, opened 2025-05-20
avr-gcc size report

charlesnicholson:
This PR adds avr-gcc (8-bit mcu) to the size reports listing.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,3 +4,7 @@ root = true
 indent_type = space
 indent_size = 2
 trim_trailing_whitespace = true
+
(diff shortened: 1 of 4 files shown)


charlesnicholson:
@stefano-zanotti-88 check it out- https://github.com/charlesnicholson/nanoprintf/actions/runs/15149070316/job/42591575692?pr=334 has the size reports. I'm not specifying an MCU though, do you have any recommendations for compiler flags? (e.g. `-mcpu=cortex-m4` gives you m4 ABI + instruction set for arm-gcc)

PR #335, opened 2025-05-22
Cosmetic changes for uptrdiff and ssize

stefano-zanotti-88:
This is symmetrical to the "a signed equivalent of size_t is needed but never defined by the standard" issue.
It also "fixes" a writeback done as unsigned, whereas it should be signed -- not that it should ever matter in practice.


charlesnicholson (approved):
Thanks!

PR #337, opened 2025-05-23
Clean up macros

stefano-zanotti-88:
This changes some `NANOPRINTF` macros to `NPF`, so that now a macro is a user-customizable option iff it is a `NANOPRINTF` one, whereas all the `NPF` ones are for private use (or guards). Of course, functions/types/etc. are still prefixed with `npf`.

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -3,8 +3,8 @@
    charles.nicholson+nanoprintf@gmail.com
    dual-licensed under 0bsd and unlicense, take your pick. see eof for details. */
 
-#ifndef NANOPRINTF_H_INCLUDED
-#define NANOPRINTF_H_INCLUDED
+#ifndef NPF_H_INCLUDED
+#define NPF_H_INCLUDED
 
 #include <stdarg.h>
 #include <stddef.h>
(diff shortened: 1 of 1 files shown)


charlesnicholson:
I think in overhauling the testing infrastructure, I will rewrite the paland tests without copying them, so all of this will live in one repo without having to tiptoe around the incompatible licenses. This is a good PR, thanks for doing it; I'll copy it over into a new PR that I do and link to this one, so that I can do the paland fixup.

charlesnicholson:
Implemented in https://github.com/charlesnicholson/nanoprintf/pull/339

stefano-zanotti-88:
Regarding paland, maybe you want to rewrite a similar test suite, or maybe you might use some or all of the "exhaustive testing" and/or "fuzz testing" I've proposed here: #343.
Note that "exhaustive" means testing all possible combinations of specifiers/options, so as to catch edge cases, but it does not mean checking other things that are more important -- like using multiple specifiers in the same call, or various cases of float precision, or cases that are known to be tricky for proper rounding, etc.
So, a lot of hand-crafted tests are still needed.
Having all these interesting cases combined with all possible combinations of options would be infeasible -- there could easily be billions of test cases.

PR #341, opened 2025-05-26
Reworked spec parsing

stefano-zanotti-88:
This PR builds on the discussion in #340.
npf_parse_format_spec() now parses whole spans of literal characters, which also includes the case of malformed conversion specifications, which start with '%' but couldn't be properly parsed because of errors or disabled-in-NPF options. This should more performant, as literal spans only require a single call, and can then be output all at once (in a loop with putc, or in a single call with puts, see #287). Also, it's easier to handle by the caller -- either NPF's printf, or the user, if this function is publicly accessible -- because such spans are treated atomically. In particular, a wrong conversion specification can be easily handled as "print everything" or "print everything except %" or "print nothing" or "print a replacement string/char". The caller only needs to check whether `conv_spec` is `NONE`, and whether the literal span (ie the string passed in for parsing) starts with '%'.

One pending issue: I've zeroed out npf_format_spec_t at the beginning. This makes for nicer handling of errors by the caller -- all the fields will be meaningful, not just `conv_spec`. Some zeroing was done before as well. What I did might even result in smaller code (it can be a memcpy/loop rather than targeted writes to multiple fields), but that should be verified.
Maybe you'll decide that the caller doesn't care about anything except `conv_spec`, in case of error/literal, in which case you can restore the previous zeroing code.


stefano-zanotti-88:
Yes, the size penalty is high, it's not worth it.
I trust your CI size reports more than my local measurements; and they cover more architectures, and most of them show a size increase.

PR #360, opened 2026-06-14
Shrink size roughtly 20%: General updates to the ASCII/FLOAT/INTEGER formatters and buffer handling

fran-cap:
**FC Notes:**
This returns around a 15 to 20 % reduction in size (Os/O2/O3 tested across a few compilers gcc13/15/clang21). 
Passes all conformance tests and was also ran against qemu platforms as well (arm m0/m4, risc-v, aarch64, ppc, and mips). 
Target was for a cortex bootloader that I reaaaaaaaaaly needed to get the size below 2k with all the bells and whistles on, it just happened to work across all archs I tested too :).
Please let me know if you have any questions or issues with any of the changes therein.

---
### Test results including new div free mode (x86 machine g++ 13.1.0):
```
  # ── Conformance suite: 384 flag combos × {C, C++} = 768 combos ──────────────

  # Default mode (division-free OFF)
  $ python3 _sizetest/gen_conformance_gcc.py
  $ ( cd tests/generated && mingw32-make -j12 )
  PASSED: 433760 assertions across 768 combos (384 flags x 2 langs)

  # Division-free mode (NANOPRINTF_USE_DIVISION_FREE_CONVERSION=1)
  $ python3 _sizetest/gen_conformance_gcc.py \
        --extra-cflags="-DNANOPRINTF_USE_DIVISION_FREE_CONVERSION=1"
  $ ( cd tests/generated && mingw32-make clean && mingw32-make -j12 )
  PASSED: 433760 assertions across 768 combos (384 flags x 2 langs)

  # ── Unit tests (doctest), LARGE=0 and LARGE=1 ──────────────────────────────

  # Default mode
  $ bash _sizetest/check_tests_gcc.sh
  [doctest] assertions: 932 | 932 passed | 0 failed |
  [doctest] assertions: 956 | 956 passed | 0 failed |

  # Division-free mode
  $ EXTRA_DEFS=-DNANOPRINTF_USE_DIVISION_FREE_CONVERSION=1 bash _sizetest/check_tests_gcc.sh
  [doctest] assertions: 932 | 932 passed | 0 failed |
  [doctest] assertions: 956 | 956 passed | 0 failed |

```
---
### Size results:
Configurations ran:
```
  - default_explicit: stock default config.
  - nofw: default minus FIELD_WIDTH.
  - nosmall: default minus SMALL length modifiers
  - noalt: default minus the ALT_FORM.
  - single: default but FLOAT_SINGLE_PRECISION.
  - lean: float-only minimum, just PRECISION + FLOAT
  - full: everything on (FIELD_WIDTH, PRECISION, FLOAT, LARGE, SMALL, BINARY, WRITEBACK, ALT_FORM, FLOAT_HEX), dp floats.
```
Results on a cortex m4 (gcc 15.2):
| Config           | m0 linked              | m4 linked              | 
|------------------|------------------------|------------------------|
| default_explicit | 2799 -> 1827 (−972)     | 2531 -> 1755 (−776)     |
| nofw             | 2131 -> 1507 (−624)     | 1959 -> 1483 (−476)     |
| nosmall          | 2751 -> 1771 (−980)     | 2423 -> 1687 (−736)     |
| noalt            | 2515 -> 1687 (−828)     | 2295 -> 1643 (−652)     | 
| single           | 2991 -> 2031 (−960)     | 2639 -> 1875 (−764)     |
| lean             | 2107 -> 1579 (−528)     | 1735 -> 1359 (−376)     | 
| full             | 4207 -> 2951 (−1256)    | 4187 -> 2687 (−1500)    | 
---
### Detailed changes:

1. Add division free conversion flag NANOPRINTF_USE_DIVISION_FREE_CONVERSAION to elimiate divsion calls for very slow micros without a mul/div block (m0).

2. Add NPF_UINT_IS_WIDE which tells the code path if the unsigned type is > 32 bits so we can take a fast path if its < 32

3. Move to using an end-pointer instead of tracking length, this frees us up to just track the returned cursor

4. Change the flags loop to use a for(;;) loop with a ++ on the switch pointer so we can force the \0 to hit the default case.

5. Combine the old >= <= digital test to a sub '0' then < 10 which saves a compare operations.

6. Refactor the conversion specifier to use a lookup table. This lets us remove parts of the lookup table with ifdefs. The benefit is by using a static table we can just handle most operations in one pass by doing the indexing off of the char type [see the (c | 32) = 'a') logic]

7. Change the utaoa_rev to return the end pointer, removed need for local counter, added npof_utoa_rev facace to give methods way to get length and keep bward compat. This saves some register spills.

8. Per .1 for integer to ASCII conversions if we dont have a divison unit and the ...DIV_FREE_CONVERSION flag is on handle the divison with multiples and shifts. Extended to work with 64 bit values as well.

9. For .1/.8 updated npf_div10 to work with just bitshifts/adds which saves space on units with no div.

10. Create handler macros (npr_bin_shr/ npf_bin_shl) for really restricted micros like cortex M0 to do single bit shifts. This is only when they're larger than 64 bit terms, on all other archs this is a normal >> <<.

11. Pack NaN InF into a single if check and return string that we just offset into.

12. Move the IEEE carry bit computation to an unconditional computation and merged the base scaling into a single expression to save a branch operation and some math.

13. For the scaling loop on division free configs handled the man_i /= 5 similar to .8 by moving to a mathamatial identify of (2*N/10 + n%10 >= 5) which lets us use the fast div10 path without having to pull in the software divider on limited platforms.

14. Change integer parth check to be a simple capacity check (end > BUIFFER_SIZE -10) since on a single point arch (32 bit) the amount of digits should be <= 10. For dual float/wide builds we keep the older path for true double conformance.

15. Clean up rounding loop to just read in the buff to a local value then check against a series of if/else terms. Similarly for the error and bail path we use a packed literal and a for loop to handle that more efficently.

16. Refactor formatter to move to a local npf_n and two macros (PUTC and PUT) which let me move the counter into a smaller loop

17. Refactor dispach to use range checks vs the old switch. This ended up being cheaper after a lot of testing to group it correctly than the old switch/jump table.

18. Change the base value calculation to happen where we call the % argument to set it to 16,10,8, etc.

19. Remove need for a switch case in NPF_EXRACT by allowing the L/UL case fall thru on targets where long is int.

20. Collapsed the default precision and '0' flag checks into enum range checks. The indexes are in order in the enums so we can just check the end for the float type or precision for d i o u etc.

21. Combined the places where we were calculating the sign, 0x hex val, and zero pad values into a single path where we calculate the pad width, fold in the zero padding/0x padding, and then check the guards for the left/right padding in one go. This lets me remove a bunch of extra checks across branches and simplifies the logic.

22. Since the string length pass aready has the len checked for  0 when the ptr is null we can remove the && cbuf check.

23. Removed cur field from npf_butfputc_ctx_t, moved to having dst be the cursor and len being the capacity that counts down.

24. Refactorted npf_butputc to remove need for bufputc_nop by moving to an if check for the dst and len being valid. A  null dst now means we're only trying to count.

25. Cleaned up logic in npf_vsnprintf to use a multi-init for the buffer and size. This was enabled since .24 lets us have a count only mechanism.

### DIFF
--- a/nanoprintf.h
+++ b/nanoprintf.h
@@ -133,12 +133,8 @@ NPF_VISIBILITY int npf_vpprintf(npf_putc pc,
   #define NANOPRINTF_USE_FLOAT_HEX_FORMAT_SPECIFIER 0
 #endif
 
-// Optional flag, defaults to 0 if not explicitly configured. On cores without a
-// hardware divider (e.g. ARMv6-M, or RISC-V without the M extension), libgcc's
-// __udivsi3 costs ~270 bytes; division-free digit extraction (shift/mask for
-// octal/hex, exact shift-add divide-by-10 for decimal) keeps it out of the link
-// entirely. Enable it explicitly (define to 1) on such targets; leave it off
-// where a hardware divider makes real division smaller.
+// Optional flag, defaults to 0 if not explicitly configured. Extracts digits
+// without integer division; smaller on cores without a hardware divider.
 #ifndef NANOPRINTF_USE_DIVISION_FREE_CONVERSION
   #define NANOPRINTF_USE_DIVISION_FREE_CONVERSION 0
 #endif
@@ -363,6 +359,33 @@ enum {
 #endif
 };
 
+// Assert range order/comparisons to enforce C standard ordering
+#define NPF_CONV_ORDER_ASSERT(NAME, COND) \
+  typedef char npf_assert_##NAME[(COND) ? 1 : -1]
+NPF_CONV_ORDER_ASSERT(text_convs_before_numeric,
+  (NPF_FMT_SPEC_CONV_PERCENT < NPF_FMT_SPEC_CONV_SIGNED_INT) &&
+  (NPF_FMT_SPEC_CONV_CHAR < NPF_FMT_SPEC_CONV_SIGNED_INT) &&
+  (NPF_FMT_SPEC_CONV_STRING < NPF_FMT_SPEC_CONV_SIGNED_INT));
+#if NANOPRINTF_USE_BINARY_FORMAT_SPECIFIERS == 1
+NPF_CONV_ORDER_ASSERT(int_convs_contiguous,
+  NPF_FMT_SPEC_CONV_UNSIGNED_INT == NPF_FMT_SPEC_CONV_SIGNED_INT + 4);
+#else
+NPF_CONV_ORDER_ASSERT(int_convs_contiguous,
+  NPF_FMT_SPEC_CONV_UNSIGNED_INT == NPF_FMT_SPEC_CONV_SIGNED_INT + 3);
+#endif
+NPF_CONV_ORDER_ASSERT(pointer_after_int_convs,
+  NPF_FMT_SPEC_CONV_POINTER > NPF_FMT_SPEC_CONV_UNSIGNED_INT);
+#if NANOPRINTF_USE_FLOAT_FORMAT_SPECIFIERS == 1
+#if NANOPRINTF_USE_WRITEBACK_FORMAT_SPECIFIERS == 1
+NPF_CONV_ORDER_ASSERT(float_convs_last,
+  NPF_FMT_SPEC_CONV_FLOAT_DEC > NPF_FMT_SPEC_CONV_WRITEBACK);
+#else
+NPF_CONV_ORDER_ASSERT(float_convs_last,
+  NPF_FMT_SPEC_CONV_FLOAT_DEC > NPF_FMT_SPEC_CONV_POINTER);
+#endif
+#endif
+#undef NPF_CONV_ORDER_ASSERT
+
 typedef struct npf_format_spec {
 #if NANOPRINTF_USE_FIELD_WIDTH_FORMAT_SPECIFIERS == 1
   int field_width;
@@ -574,7 +597,7 @@ static NPF_FORCE_INLINE int npf_parse_format_spec(char const *format,
 }
 
 #if NANOPRINTF_USE_DIVISION_FREE_CONVERSION == 1
-// Exact n/10 for any uint32_t without hardware division (Hacker's Delight).
+// Calculate div by 10 justing a shift and mask to avoid using hw div operands
 static NPF_NOINLINE uint32_t npf_div10(uint32_t n) {
   uint32_t q = (n >> 1) + (n >> 2);
   q += q >> 4;
@@ -589,9 +612,7 @@ static NPF_NOINLINE char *npf_utoa_rev_end(
     npf_uint_t val, char *buf, uint_fast8_t base, char case_adj) {
 #if (NANOPRINTF_USE_LARGE_FORMAT_SPECIFIERS == 1) || \
     ((NANOPRINTF_USE_DIVISION_FREE_CONVERSION == 1) && NPF_UINT_IS_WIDE)
-  // For values that don't fit in 32 bits, shift-and-subtract divmod avoids
-  // pulling in __udivmoddi4 (~784 B on ARM EABI). Slow but tiny; the 32-bit
-  // fast path below handles all typical values.
+  // Use shift and subtract here to avoid hw div operation
   while (val > 0xFFFFFFFFu) {
     npf_uint_t q = 0, r = 0;
     for (int i = (int)(sizeof(val) * 8) - 1; i >= 0; --i) {
@@ -701,9 +722,8 @@ static NPF_FORCE_INLINE npf_real_bin_t npf_real_to_int_rep(npf_real_t f) {
 }
 
 #if NANOPRINTF_USE_DIVISION_FREE_CONVERSION == 1
-// Variable shifts of 64-bit values call __aeabi_llsl/__aeabi_llsr on cores
-// without 64-bit shifters (e.g. ARMv6-M); one-bit-at-a-time loops keep those
-// helpers out of the link. 32-bit npf_real_bin_t shifts stay native.
+// Variable shifts of 64-bit values call sw helpers on archs
+// without 64-bit shifters. Perfer 1 bit shits to keep in 32 bit word spce.
 static NPF_FORCE_INLINE npf_real_bin_t npf_bin_shr(npf_real_bin_t v, int_fast8_t s) {
   if (sizeof(v) > sizeof(uint32_t)) { while (s--) { v >>= 1; } return v; }
   return (npf_real_bin_t)(v >> s);
@@ -720,6 +740,8 @@ static NPF_FORCE_INLINE npf_real_bin_t npf_bin_shl(npf_real_bin_t v, int_fast8_t
 #endif
 
 static int npf_ftoa_rev(char *buf, npf_format_spec_t const *spec, npf_real_t f) {
+  // Packed reversed specials "NAN"/"FNI"(inf)/"RRE"(err), selected by offset.
+  static char const specials[] = "NAN\0FNI\0RRE";
   char const *ret = NULL;
   npf_real_bin_t bin = npf_real_to_int_rep(f);
 
@@ -729,9 +751,7 @@ static int npf_ftoa_rev(char *buf, npf_format_spec_t const *spec, npf_real_t f) 
 
   bin &= ((npf_real_bin_t)0x1 << NPF_REAL_MAN_BITS) - 1;
   if (!((unsigned)(exp + 1) & NPF_REAL_EXP_MASK)) { // special value
-    // Packed reversed specials "NAN"/"FNI"(inf)/"RRE"(err), indexed by offset.
-    // Cast this so we dont throw a warning for the + operation
-    ret = (char const *)"NAN\0FNI\0RRE" + ((bin) ? 0 : 4);
+    ret = specials + (bin ? 0 : 4);
     goto exit;
   }
   if (spec->prec > (NANOPRINTF_CONVERSION_BUFFER_SIZE - 2)) { goto exit; }
@@ -890,7 +910,7 @@ static int npf_ftoa_rev(char *buf, npf_format_spec_t const *spec, npf_real_t f) 
 
   return (int)end;
 exit:
-  if (!ret) { ret = (char const *)"NAN\0FNI\0RRE" + 8; }
+  if (!ret) { ret = specials + 8; }
   uint_fast8_t i = 0;
   do { buf[i] = (char)(ret[i] + spec->case_adjust); } while (ret[++i]);
   return -(int)i;
@@ -1024,13 +1044,12 @@ static void npf_bufputc(int c, void *ctx) {
 }
 
 #define NPF_PUTC(VAL) do { pc((int)(VAL), pc_ctx); ++npf_n; } while (0)
-// Emission inside a conversion: the spec's total length is added to npf_n in bulk.
 #define NPF_PUT(VAL) do { pc((int)(VAL), pc_ctx); } while (0)
 
 #define NPF_EXTRACT(DST, MOD, CAST_TO, EXTRACT_AS) \
   case NPF_FMT_SPEC_LEN_MOD_##MOD: DST = (CAST_TO)va_arg(args, EXTRACT_AS); break
 
-// When sizeof(long) == sizeof(int) (any ILP32 ABI, e.g. ARM EABI / Win32 / Win64),
+// When sizeof(long) == sizeof(int)
 // va_arg(*args, long) and va_arg(*args, int) read the same bits, so the LONG
 // case can fall into the int-promotion default.
 #if LONG_MAX == INT_MAX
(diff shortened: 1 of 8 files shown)


fran-cap:
Just as a note the build for the windows build is broken submitted #361 to fix this.

fran-cap:
Thanks for accepting my PR for 621 , just rebasing so that this should be able to show a clean/check merge, apologies for the noise.

charlesnicholson:
Haha yes, the first one was the easy one :) I'll look at the big size reduction ASAP, thanks for opening it, I'm excited to study it!

charlesnicholson:
Hi @fran-cap - thanks for this PR; I'm very excited to review it! Sorry, life is crazy right now but I hope to get to it soon.

nanoprintf.h
```diff
@@ -498,85 +512,124 @@ static int npf_parse_format_spec(char const *format, npf_format_spec_t *out_spec
...
+      NPF_FMT_SPEC_CONV_OCTAL,           // 'o'
+      NPF_FMT_SPEC_CONV_POINTER,         // 'p'
+      0, 0,                              // 'q', 'r'
+      NPF_FMT_SPEC_CONV_STRING,          // 's'
+      0,                                 // 't'
+      NPF_FMT_SPEC_CONV_UNSIGNED_INT,    // 'u'
+      0, 0,                              // 'v', 'w'
+      NPF_FMT_SPEC_CONV_HEX_INT,         // 'x'
+    };
+    unsigned const idx = (unsigned)((c | 32) - 'a');
```
charlesnicholson, line 583: Before this change, main would print the literal format specifiers if they were invalid (e.g. if you had `%P` in your format string, npf would print `%P`), but now they're implicitly all made uppercase.

Since it's all UB anyway, I don't mind, but it would be nice to explicitly mention this in the docs. Would you mind updating the docs to explain this to users please? I don't mind if it breaks anyone since they were depending on bad UB, but it'd be nice to inform users.

nanoprintf.h
```diff
@@ -921,43 +1019,40 @@ static int npf_bin_len(npf_uint_t u) {
...
+#else
+  #define NPF_LONG_IS_INT 0
+#endif
 
-#define NPF_EXTRACT(MOD, CAST_TO, EXTRACT_AS) \
-  case NPF_FMT_SPEC_LEN_MOD_##MOD: val = (CAST_TO)va_arg(args, EXTRACT_AS); break
 
 #define NPF_WRITEBACK(MOD, TYPE) \
-  case NPF_FMT_SPEC_LEN_MOD_##MOD: *(va_arg(args, TYPE *)) = (TYPE)pc_cnt.n; break
+  case NPF_FMT_SPEC_LEN_MOD_##MOD: *(va_arg(args, TYPE *)) = (TYPE)npf_n; break
```
charlesnicholson: you've changed the implementation of the `NPF_WRITEBACK` macro but also removed all of its uses. I suspect that the change was mechanical, so could I ask you to simply delete the macro definition entirely please?

nanoprintf.h
```diff
@@ -978,47 +1073,36 @@ int npf_vpprintf(npf_putc pc, void *pc_ctx, char const *format, va_list args) {
...
-          break;
-        default:
-          break;
-      }
-    }
+    // For d i o u x X, the '0' flag must be ignored if a precision is provided.
+    // Those conversions are contiguous in the enum (except BINARY, b).
+    if ((fs.prec_opt != NPF_FMT_SPEC_OPT_NONE) &&
+        (fs.conv_spec >= NPF_FMT_SPEC_CONV_SIGNED_INT) &&
+        (fs.conv_spec <= NPF_FMT_SPEC_CONV_UNSIGNED_INT)
```
charlesnicholson, line 1116: Changing this to a range comparison is fine, but I'd really like to see some static assertions in place here, in case they ever get changed or reordered in a way that breaks things subtly.

nanoprintf.h
```diff
@@ -133,6 +133,16 @@ NPF_VISIBILITY int npf_vpprintf(npf_putc pc,
   #define NANOPRINTF_USE_FLOAT_HEX_FORMAT_SPECIFIER 0
 #endif
 
+// Optional flag, defaults to 0 if not explicitly configured. On cores without a
+// hardware divider (e.g. ARMv6-M, or RISC-V without the M extension), libgcc's
+// __udivsi3 costs ~270 bytes; division-free digit extraction (shift/mask for
+// octal/hex, exact shift-add divide-by-10 for decimal) keeps it out of the link
+// entirely. Enable it explicitly (define to 1) on such targets; leave it off
+// where a hardware divider makes real division smaller.
```
charlesnicholson: I'd shorten this comment, remove the example architectures, remove the libgcc references, and move the implementation comment to the implementation. (This feels like an AI LLM comment; they're not great at comment scope yet)

nanoprintf.h
```diff
@@ -133,6 +133,16 @@ NPF_VISIBILITY int npf_vpprintf(npf_putc pc,
...
+#ifndef NANOPRINTF_USE_DIVISION_FREE_CONVERSION
+  #define NANOPRINTF_USE_DIVISION_FREE_CONVERSION 0
+#endif
```
charlesnicholson, line 140: This needs to be added to the test matrix, to make sure that we don't backslide on conformance / behavior.

charlesnicholson: Also please add this to the documentation; it's an important flag.

nanoprintf.h
```diff
@@ -921,43 +1019,40 @@ static int npf_bin_len(npf_uint_t u) {
...
 
-#define NPF_PUTC(VAL) do { npf_putc_cnt((int)(VAL), &pc_cnt); } while (0)
+// When sizeof(long) == sizeof(int) (any ILP32 ABI, e.g. ARM EABI / Win32 / Win64),
+// va_arg(*args, long) and va_arg(*args, int) read the same bits, so the LONG
+// case can fall into the int-promotion default.
+#if LONG_MAX == INT_MAX
+  #define NPF_LONG_IS_INT 1
+#else
+  #define NPF_LONG_IS_INT 0
+#endif
```
charlesnicholson, line 1059: these leak out of the header file; please undef them later

nanoprintf.h
```diff
@@ -921,43 +1019,40 @@ static int npf_bin_len(npf_uint_t u) {
...
 
-static void npf_bufputc_nop(int c, void *ctx) { (void)c; (void)ctx; }
-
-typedef struct npf_cnt_putc_ctx {
-  npf_putc pc;
-  void *ctx;
-  int n;
-} npf_cnt_putc_ctx_t;
+#define NPF_PUTC(VAL) do { pc((int)(VAL), pc_ctx); ++npf_n; } while (0)
+// Emission inside a conversion: the spec's total length is added to npf_n in bulk.
```
charlesnicholson: this is another agent / llm comment that would be more useful if moved to the actual implementation

nanoprintf.h
```diff
@@ -656,8 +728,10 @@ static int npf_ftoa_rev(char *buf, npf_format_spec_t const *spec, npf_real_t f)
     (npf_ftoa_exp_t)((npf_ftoa_exp_t)(bin >> NPF_REAL_MAN_BITS) & NPF_REAL_EXP_MASK);
 
   bin &= ((npf_real_bin_t)0x1 << NPF_REAL_MAN_BITS) - 1;
-  if (exp == (npf_ftoa_exp_t)NPF_REAL_EXP_MASK) { // special value
-    ret = (bin) ? "NAN" : "FNI";
+  if (!((unsigned)(exp + 1) & NPF_REAL_EXP_MASK)) { // special value
+    // Packed reversed specials "NAN"/"FNI"(inf)/"RRE"(err), indexed by offset.
+    // Cast this so we dont throw a warning for the + operation
+    ret = (char const *)"NAN\0FNI\0RRE" + ((bin) ? 0 : 4);
```
charlesnicholson: this string is used twice in this function, could / should be static char const * at function scope and referred to by name instead of repeated.

nanoprintf.h
```diff
@@ -647,6 +700,25 @@ static NPF_FORCE_INLINE npf_real_bin_t npf_real_to_int_rep(npf_real_t f) {
...
+}
+static NPF_FORCE_INLINE npf_real_bin_t npf_bin_shl(npf_real_bin_t v, int_fast8_t s) {
+  if (sizeof(v) > sizeof(uint32_t)) { while (s--) { v <<= 1; } return v; }
+  return (npf_real_bin_t)(v << s);
+}
+  #define NPF_BIN_SHR(V, S) npf_bin_shr((V), (int_fast8_t)(S))
+  #define NPF_BIN_SHL(V, S) npf_bin_shl((V), (int_fast8_t)(S))
+#else
+  #define NPF_BIN_SHR(V, S) ((npf_real_bin_t)((V) >> (S)))
+  #define NPF_BIN_SHL(V, S) ((npf_real_bin_t)((V) << (S)))
```
charlesnicholson, line 739: these should be undef'd at the end of the file; we don't want to leak out any NPF_ internal macros.

charlesnicholson (approved):
This looks great, I left some feedback that needs addressing. Thanks for the effort! Also I'm curious; what agent / setting did you use? I haven't pointed Fable at nanoprintf yet, but I tried Sonnet / Haiku back in the day and they did not do too well.

charlesnicholson:
Sorry for the delay, @fran-cap - I think the changes overall are great! If you don't mind addressing the feedback, I'd be happy to merge it.

fran-cap:
Sure thing, this is some good fb. Give me a day to review these and I'll submit another push.

nanoprintf.h
fran-cap: Added a short note to the readme about this.

nanoprintf.h
fran-cap: Concur, removed completely.

nanoprintf.h
fran-cap: Agreed, I ended up making a NPF_CONV_ORDER_ASSERT macro then having a short check block. Let me know if thats preferred or if you want to use the _Static_assert/static_assert style.

nanoprintf.h
fran-cap: Concur, I always go through my agent code and re-write the comments, this just slipped thru.

nanoprintf.h
fran-cap: Done, added to the readme and implmented a new unit test block for this *divfree.cc, also added to the makefile/python test script.

nanoprintf.h
fran-cap: Agreed, undefed at the end, too used to writing in C files where I can get away with that ;)

nanoprintf.h
fran-cap: Agreed, moved.

nanoprintf.h
fran-cap: Yeah I went back and forth on this a few times, moved this to a specials[] const char block now.

nanoprintf.h
fran-cap: Agreed, same comment as above, need to remember I'm working with a .h file, added undefs.

fran-cap:
@charlesnicholson reviewed, and just pushed up w/ comments, I agree w/ all the requested changes. As far as your above question about agent approach I use a mixture of opus 4.8 and gpt 5.4 teams with me writing orchestration files, most of the stuff I do for work is low level optimization so I have extensive internal documentation that I use for cross reference.

Overall the process for this push was around 2 weeks of that (maybe 30-50mTok total) with me hand reviewing the changes to make edits for style and general format. 
'Course a few stray comments slip though ;) 

I find that having previous work then testing against that work gets the best results as long as you're willing to have a good methodology of tracking every single test/change over an extended period of time. I find that having all the agents work off of a single shared knowledge document helps keep them focused and provides better results than using a git history or worktree style, at least for a single focused project like this, for a larger project the worktree approach is better.

Of course the models do make tons of mistakes when it comes to style/approach but as long as you're willing to review and read the code I find the overall approach effective for long micro-optimization tasks like this.

charlesnicholson (approved):

charlesnicholson:
That's more or less how i'm working with these things too - I threw Fable at a few of our internal systems at work and it did amazing things; it needed some micromanagement and comment editing but I'm so impressed by how far these things have come! I was going to point Fable at nanoprintf but then the pricing changed and I can't quite justify the cost on my employers :)

Anyway, thanks for the work, it's a really great set of improvements to nanoprintf and I'm happy to merge them. I'll cut a release after I get some mileage on it, assuming nothing goes wrong.

charlesnicholson:
Oh, btw, @fran-cap - please add yourself to the contributors file! Once you do that (or let me know that you don't want to), I'll merge

fran-cap:
Oop that main merge made it a bit awkward to stage but have that pushed up @charlesnicholson , if you'd want to cherrypick that or have me squash reset lemme know.

charlesnicholson:
Ah, sorry @fran-cap I didn't realize that would make things harder for you. If the only change is adding your name to the contributors, you might be able to sneak that in via the web UI? Or I can merge this and we can do a fast-follow commit that just has your name, or you can force-push over my merge, or whatever. Sorry about that; please do whatever you want / need to fix it.

fran-cap:
Not a problem can merge in as is @charlesnicholson then do the fast follow afterwards.

PR #361, opened 2026-06-14
Fix windows CI path to work with the latest windows git contianer

fran-cap:
Git updated and the container for the windows builds no longer works. 
Simple 1 liner to locate MSVC via vswhere instead of  via the old hardcoded VS 2022 Enterprise path.
Old builds were using Image: windows-2022 but the latest image now has Image: windows-2025-vs2026 so the vs is on a different path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,7 +143,8 @@ jobs:
       - name: Build
         shell: cmd
         run: |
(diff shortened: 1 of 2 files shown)


charlesnicholson (approved):

charlesnicholson:
Thanks for the fix!

PR #366, opened 2026-07-13
Document npf_putc callbacks

### LINKED ISSUE #241: Better documentation for putc callback
The readme doesn't give an example of the actual npf_putc callback, just how to call nanoprintf with a callback.

Add an actual callback, with use of the context pointer, to the readme to make it clearer.

aouxwoux:
Replace the undefined `my_uart_putc` placeholder with a bounded `npf_putc` callback that uses the context pointer to keep track of output.

Also add a small variadic wrapper that forwards its `va_list` to `npf_vpprintf`, as requested in the issue.

Tests:
- `git diff --check`
- Not run: a C compiler is unavailable in this environment.

Fixes #241.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,14 +32,21 @@ Then, in any file where you want to use nanoprintf, simply include the header an
 ```
 #include "nanoprintf.h"
 
+static void my_uart_putc(int c, void *ctx) {
+  uart_write((my_uart_t *)ctx, c);
+}
+
 void print_to_uart(void) {
-  npf_pprintf(&my_uart_putc, NULL, "Hello %s%c %d %u %f\n", "worl", 'd', 1, 2, 3.f);
+  my_uart_t *uart = get_my_uart();
+  npf_pprintf(&my_uart_putc, uart, "Hello %s%c %d %u %f\n", "worl", 'd', 1, 2, 3.f);
 }
 
 void print_to_buf(void *buf, unsigned len) {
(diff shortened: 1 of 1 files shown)


charlesnicholson:
I'm not crazy about this; it's less about how the nanoprintf putc callback works and more an example of how to reinvent snprintf through the nanoprintf API. If you wanted to change
```
#include "nanoprintf.h"

void print_to_uart(void) {
  npf_pprintf(&my_uart_putc, NULL, "Hello %s%c %d %u %f\n", "worl", 'd', 1, 2, 3.f);
}

void print_to_buf(void *buf, unsigned len) {
  npf_snprintf(buf, len, "Hello %s", "world");
}
```

to something like
```
#include "nanoprintf.h"

static void my_uart_putc(int c, void *ctx) {
  uart_write((my_uart_t *)ctx, c);
}

void print_to_uart(void) {
  my_uart_t *uart = get_my_uart();
  npf_pprintf(&my_uart_putc, uart, "Hello %s%c %d %u %f\n", "worl", 'd', 1, 2, 3.f);
}

void print_to_buf(void *buf, unsigned len) {
  npf_snprintf(buf, len, "Hello %s", "world");
}
```

I think that would be a better improvement. As it is, your PR adds a bunch of lines of code that the user has to study, that are more about implementing a bounds-checked snprintf than they are about how the nanoprintf putc stuff actually works.

charlesnicholson (approved):

